Issue #8, opened 2025-12-10
Develop a tutorial for doing local ansible development with boxman


mherkazandjian:
assignees: @Orski174 @mherkazandjian

- markdown, github
- target audiance: someone with basic knowledge of linux and sys admin
- requirements: low resources laptop: 8GB ram, 50GB free storage, ease of use is paramount
- components of the persentation
  -  chapter 1
     - introduction
     - add a demo (screen recording)
     - glance over the config
     - how to install libvirt and its dependencies
     - consolidate the location of the image location
     - add instructions on how to import the image into libvirt (maybe add a wrapper)
     - spin up a one machine cluster
     - run a simple ansible playbook with three tasks
       - ping the host
       - install a package
       - change a line to a configuration file
  - chapter 2
    - introduction to ansible
    - describe how an collection looks like
    - how to set up an ansible collection
      - set up an ansible collections in default paths (not ideal for prod)
      - set up ansible collections in custom dirs (for dev)
    - .. todo:: add more items as we go

PR #9, opened 2025-12-10
placeholder readme for the tutorial

mherkazandjian:
see issue-#8


mherkazandjian:
wont't do , we decided to use latex based for the time being to experiement.

Issue #6, opened 2025-05-15
allow specifying some or all of  ssh_config when provisioning the machines and even overriding the default one


mherkazandjian:
```bash
 Host *
    StrictHostKeyChecking no
    UserKnownHostsFile /dev/null
    ServerAliveInterval 30
    ForwardAgent yes
    ForwardX11Timeout 596h
    Compression yes
    ForwardX11 no
    SendEnv GIT_*

.
.
.
```

Issue #5, opened 2025-05-12
support setting a unique address / network if the {{ unique }} or something of that sort TBD is specified


mherkazandjian:
![Image](https://github.com/user-attachments/assets/edcc37b4-2499-4c93-b095-a30f10e5c015)

PR #59, opened 2026-07-13
feat(dc-provider): Phase 1 — provider registry & per-cluster dispatch (no behavior change)

Orski174:
Closes #48's successor phase — Phase 1 of epic #42. Targets the epic feature branch `feat/docker-compose-provider`. Plan: [#49 comment](https://github.com/mherkazandjian/boxman/issues/49#issuecomment-4959931644).

## What

Three commits, reviewable independently:

1. **`f90202c` registry + manager surface** — `boxman.providers` gains `PROVIDERS` (lazy factories: libvirt / virtualbox / docker-compose), `create_session()` (unknown type → `ValueError` listing supported; docker-compose → friendly "lands in Phase 3 #51"), `primary_provider_type()`. `BoxmanManager` gains `register_session()`, `provider_type_for_cluster()` (per-cluster `provider:` key wins — official in Phase 2), `session_for_cluster()`, `session_for_vm()` (full-name resolution via `_vm_cluster_map()`, default-session fallback), and the legacy `provider` property becomes a typed compat shim whose setter keeps `manager.provider = mock` tests feeding threaded flows. New `tests/test_provider_registry.py` (13 tests incl. the two-cluster shared-session AC).
2. **`ee7e9f1` app wiring** — every key under `provider:` gets a registry-built session; libvirt keeps its config-enrichment verbatim; import-image keeps its manifest-driven single-provider flow.
3. **`59e9f6c` flow threading** — pure call-site substitution (no tuple shapes changed): networks, clone/destroy, configure+start, IP/ssh/connect, provision retry/eject, up/down, snapshot list/log/take/restore/delete, storage compress, control verbs, and all `update_provider_config_with_runtime` preambles (now applied to every registered session).

## Side effects flagged for review

- **Latent bug fixed**: the old virtualbox branch created a session and never attached it (`manager.provider` stayed `None`); uniform registration fixes that.
- Registry errors now log cleanly and exit 2 instead of tracebacking / silently proceeding sessionless.

## Deliberately kept on the default session (each marked `# Phase 1 (#49)` in code)

import-image (single-provider by manifest) · template management + storage pools (libvirt, project-level) · VMs removed from the config (cluster unresolvable) · snapshot collapse (libvirt qcow2 chains) · update/diff + PXE flows (libvirt-shaped; only libvirt clusters can exist until Phase 3 #51).

## Verification

- Full suite **1119 passed** on the host (py3.12) **and inside the epic staging VM** (`boxes/dc-provider-staging`, branch checked out at `59e9f6c`)
- Live smoke against the real staging project on this host: threaded `snapshot list` resolves per-cluster and finds the `pristine` snapshot; `boxman conf` clean
- v1.0 behavior: byte-identical paths — all provider calls route through the same session object as before (single-provider configs resolve every cluster to it)

## Drift from the issue text (also going in the retro)

- "split into small PRs per flow" → **one PR, three commits** (single review round per the epic flow)
- The update/diff + PXE flows grew after the issue was written; they stay on the shim until Phase 3 (see keep-list)

### DIFF
--- a/src/boxman/manager.py
+++ b/src/boxman/manager.py
@@ -50,7 +50,15 @@ class BoxmanManager:
         #: Dict[str, ProviderSession]: live sessions keyed by provider
         #: type ('libvirt', 'virtualbox', …); populated by
         #: :meth:`register_session` and consumed by
-        #: :meth:`session_for_cluster`
+        #: :meth:`session_for_cluster`.
+        #:
+        #: .. note:: Keying by provider *type* is correct for Phase 1:
+        #:    a libvirt host has one connection, so two libvirt clusters
+        #:    share one session. The per-cluster session model ADR-001
+        #:    needs (one docker-compose *project* per cluster) re-keys
+        #:    this map by cluster in Phase 3 (#51); the
+        #:    :meth:`session_for_cluster` / :meth:`session_for_vm` seam
+        #:    keeps every call site unchanged when that lands.
         self._sessions: dict[str, ProviderSession] = {}
 
         #: the logger instance
@@ -116,7 +124,11 @@ class BoxmanManager:
         """
         self._provider = value
         if value is not None:
-            self._get_sessions()[primary_provider_type(self.config)] = value
+            # ``getattr`` keeps the setter safe on ``__new__``-built
+            # managers (used in tests), which have no ``config`` yet.
+            self._get_sessions()[
+                primary_provider_type(getattr(self, 'config', None))
+            ] = value
 
     def _get_sessions(self) -> dict[str, "ProviderSession"]:
         """
@@ -181,14 +193,22 @@ class BoxmanManager:
         """
         provider_type = self.provider_type_for_cluster(cluster_name)
         session = self._get_sessions().get(provider_type)
-        if session is None:
-            raise ValueError(
-                f"no provider session registered for provider "
-                f"'{provider_type}' (needed by cluster '{cluster_name}')"
-            )
-        return session
+        if session is not None:
+            return session
+        # Compat parity: when nothing is registered for the project's
+        # *primary* type but a default session was assigned directly
+        # (older tests / call sites poke ``_provider`` without going
+        # through :meth:`register_session`), resolve to it — exactly
+        # what the old single ``self.provider`` did for every flow.
+        default = getattr(self, '_provider', None)
+        if default is not None and provider_type == primary_provider_type(self.config):
+            return default
+        raise ValueError(
+            f"no provider session registered for provider "
+            f"'{provider_type}' (needed by cluster '{cluster_name}')"
+        )
 
-    def session_for_vm(self, full_vm_name: str) -> "ProviderSession":
+    def session_for_vm(self, full_vm_name: str) -> Optional["ProviderSession"]:
         """
         Resolve the provider session that manages *full_vm_name*.
 
@@ -202,7 +222,10 @@ class BoxmanManager:
 
         Returns:
             The provider session managing the VM's cluster, or the
-            default session when the name is not in the config.
+            default session when the name is not in the config. May be
+            ``None`` when the name is unknown and no default session has
+            been registered — the same NoneType behaviour the old
+            ``self.provider`` call sites had.
         """
         cluster_name = self._vm_cluster_map().get(full_vm_name)
         if cluster_name is None:
@@ -219,12 +242,18 @@ class BoxmanManager:
         loop) can resolve the owning cluster without parsing names.
 
         Returns:
-            Mapping of full VM name to cluster name.
+            Mapping of full VM name to cluster name. Empty when the
+            config is unset or has no ``project`` (e.g. the import-image
+            provider-slice config) so :meth:`session_for_vm` cleanly
+            falls back to the default session instead of raising.
         """
-        prj_name = f'bprj__{self.config["project"]}__bprj'
+        config = self.config or {}
+        if 'project' not in config:
+            return {}
+        prj_name = f'bprj__{config["project"]}__bprj'
         return {
             f"{prj_name}_{cluster_name}_{vm_name}": cluster_name
-            for cluster_name, cluster in ((self.config or {}).get('clusters') or {}).items()
+            for cluster_name, cluster in (config.get('clusters') or {}).items()
             for vm_name in (cluster.get('vms') or {}).keys()
         }
 
@@ -1121,7 +1150,7 @@ class BoxmanManager:
 
         Designed to be used with a Cobbler PXE provisioning server.
         """
-        session = cls.provider
+        session = cls.provider  # Phase 1 (#49): single-VM PXE flow stays on the default session until Phase 3
         vm_name = cli_args.vm
 
         cls.logger.info(f"setting boot order to [network, hd] for '{vm_name}'")
@@ -1881,7 +1910,7 @@ class BoxmanManager:
                 # fallback: try virsh list
                 try:
                     from boxman.providers.libvirt.commands import VirshCommand
-                    provider_type = list(self.config.get('provider', {}).keys())[0]
+                    provider_type = primary_provider_type(self.config)
                     provider_config = self.config.get('provider', {}).get(provider_type, {})
                     if self.app_config and 'providers' in self.app_config:
                         app_prov = self.app_config['providers'].get(provider_type, {})
@@ -3178,10 +3207,7 @@ class BoxmanManager:
 
         # Resolve provider config so VirshCommand uses the right
         # runtime / URI / sudo settings.
-        provider_type = (
-            list(self.config.get('provider', {}).keys())[0]
-            if 'provider' in self.config else 'libvirt'
-        )
+        provider_type = primary_provider_type(self.config)
         provider_config = self.config.get('provider', {}).get(provider_type, {})
         if self.app_config and 'providers' in self.app_config:
             app_prov = self.app_config['providers'].get(provider_type, {})
@@ -3210,10 +3236,7 @@ class BoxmanManager:
         project = self.config.get("project", "")
         prj_prefix = f"bprj__{project}__bprj_"
 
-        provider_type = (
-            list(self.config.get('provider', {}).keys())[0]
-            if 'provider' in self.config else 'libvirt'
-        )
+        provider_type = primary_provider_type(self.config)
         provider_config = self.config.get('provider', {}).get(provider_type, {})
         if self.app_config and 'providers' in self.app_config:
             app_prov = self.app_config['providers'].get(provider_type, {})
@@ -3248,10 +3271,7 @@ class BoxmanManager:
         if not expected:
             return {}
 
-        provider_type = (
-            list(self.config.get('provider', {}).keys())[0]
-            if 'provider' in self.config else 'libvirt'
-        )
+        provider_type = primary_provider_type(self.config)
         provider_config = self.config.get('provider', {}).get(provider_type, {})
         if self.app_config and 'providers' in self.app_config:
             app_prov = self.app_config['providers'].get(provider_type, {})
@@ -4432,7 +4452,7 @@ class BoxmanManager:
         """
         from boxman.providers.libvirt.storage import vm_disk_paths
 
-        storage = cls.provider.storage
+        storage = cls.provider.storage  # Phase 1 (#49): storage_df stays on the default session until Phase 3
         rows = []
         snap_mem_total_per_vm: dict[str, int] = {}
         for full_vm_name, workdir, vm_info in cls._storage_targets(cls, cli_args):
@@ -4491,7 +4511,7 @@ class BoxmanManager:
         Warns when a VM's disks lack ``discard='unmap'`` — fstrim will succeed
         but nothing will be returned to the host.
         """
-        storage = cls.provider.storage
+        storage = cls.provider.storage  # Phase 1 (#49): storage_trim stays on the default session until Phase 3
         for full_vm_name, _workdir, _vm_info in cls._storage_targets(cls, cli_args):
             if not storage.is_running(full_vm_name):
                 cls.logger.warning(
@@ -4577,7 +4597,7 @@ class BoxmanManager:
         no_shutdown = getattr(cli_args, 'no_shutdown', False)
         dry_run = getattr(cli_args, 'dry_run', False)
 
-        provider_config = cls.provider.provider_config
+        provider_config = cls.provider.provider_config  # Phase 1 (#49): storage_compact stays on the default session until Phase 3
         processes = [
             Process(
                 target=BoxmanManager._compact_one_vm,
@@ -4892,10 +4912,7 @@ class BoxmanManager:
             return
 
         # Query virsh for states
-        provider_type = (
-            list(cls.config.get("provider", {}).keys())[0]
-            if "provider" in cls.config else "libvirt"
-        )
+        provider_type = primary_provider_type(cls.config)
         provider_config = cls.config.get("provider", {}).get(provider_type, {})
         if cls.app_config and "providers" in cls.app_config:
             app_prov = cls.app_config["providers"].get(provider_type, {})
--- a/src/boxman/manager.py
+++ b/src/boxman/manager.py
@@ -15,11 +15,12 @@ import yaml
 from boxman.utils.shell import run
 
 from boxman import log
+from boxman.abstract.providers import ProviderSession
 from boxman.config_cache import BoxmanCache
 from boxman.image_cache import ImageCache
 from boxman.netlab import ContainerlabManager, shared_bridges
+from boxman.providers import primary_provider_type
 from boxman.providers.libvirt.commands import VirshCommand
-from boxman.providers.libvirt.session import LibVirtSession
 from boxman.runtime import RuntimeBase, create_runtime
 from boxman.task_runner import TaskRunner
 from boxman.utils.io import write_files
@@ -42,8 +43,15 @@ class BoxmanManager:
         #: Optional[Dict[str, Any]]: the loaded configuration dictionary
         self.config: dict[str, Any] | None = None
 
-        #: the private backing field for the provider property
+        #: the private backing field for the provider property (the
+        #: default session — see :meth:`register_session`)
         self._provider = None
+
+        #: Dict[str, ProviderSession]: live sessions keyed by provider
+        #: type ('libvirt', 'virtualbox', …); populated by
+        #: :meth:`register_session` and consumed by
+        #: :meth:`session_for_cluster`
+        self._sessions: dict[str, ProviderSession] = {}
 
         #: the logger instance
         self.logger = log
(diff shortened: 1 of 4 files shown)


src/boxman/manager.py
```diff
@@ -78,24 +86,147 @@ def load_app_config(self, config: dict[str, Any]) -> None:
...
+        sites and tests working: the assigned session is also registered
+        under the project's primary provider type so that
+        :meth:`session_for_cluster` resolves it.
 
         Args:
             value: The provider session instance
         """
         self._provider = value
+        if value is not None:
+            self._get_sessions()[primary_provider_type(self.config)] = value
```
mherkazandjian: **[worth addressing] The setter is not `__new__`-safe — yet `__new__`-built managers are exactly its target audience.** `_get_sessions()` grew its lazy guard specifically for managers built via `BoxmanManager.__new__` (per its own docstring), but this line still dereferences `self.config`, which doesn't exist on such instances:

```python
m = BoxmanManager.__new__(BoxmanManager)
m.provider = MagicMock()
# AttributeError: 'BoxmanManager' object has no attribute 'config'
```

(Reproduced against this branch.) Every current test happens to assign `mgr.config` first — e.g. `tests/test_snapshot_selection.py` builds via `__new__`, sets config, then hits this setter — so the suite passes, but the compat contract is one attribute-order swap away from breaking. `primary_provider_type(getattr(self, 'config', None))` closes it.

src/boxman/manager.py
```diff
@@ -78,24 +86,147 @@ def load_app_config(self, config: dict[str, Any]) -> None:
...
+
+        Args:
+            cluster_name: The cluster name as it appears under
+                ``clusters:`` in the project config.
+
+        Returns:
+            The provider type name.
+        """
+        cluster = ((self.config or {}).get('clusters') or {}).get(cluster_name) or {}
+        return cluster.get('provider') or primary_provider_type(self.config)
```
mherkazandjian, line 176: **[worth addressing] The per-cluster key is honored here, but app.py never builds sessions for cluster-level types — so exercising the very feature this line (and `test_cluster_level_provider_key_wins`) advertises yields a raw mid-flow traceback instead of the friendly registry error.**

Scenario on this branch: top-level `provider: {libvirt: …}` plus `clusters: {services: {provider: docker-compose, …}}` → app.py's `provider_types` (top-level keys only) registers just libvirt → `boxman provision` defines the first cluster's networks, then dies with `ValueError: no provider session registered for provider 'docker-compose' (needed by cluster 'services')` — potentially inside a forked `Process` child, after partial provisioning. The "lands in Phase 3 #51" message never fires because `create_session('docker-compose', …)` is never called on this path.

Cheap Phase-1 fix: right after session registration in app.py, validate that `provider_type_for_cluster(c)` resolves to a registered session for every cluster (or fold cluster-level keys into `provider_types`), so this fails with the same clean exit-2 the top-level path gets.

src/boxman/manager.py
```diff
@@ -78,24 +86,147 @@ def load_app_config(self, config: dict[str, Any]) -> None:
...
+        Args:
+            full_vm_name: The full (project-prefixed) VM name.
+
+        Returns:
+            The provider session managing the VM's cluster, or the
+            default session when the name is not in the config.
+        """
+        cluster_name = self._vm_cluster_map().get(full_vm_name)
+        if cluster_name is None:
+            return self.provider
```
mherkazandjian, line 232: **[worth addressing] `manager.provider is not None` no longer implies threaded flows can resolve a session — the shim's two halves can diverge.** Anything that sets `_provider` without populating `_sessions` gets split behavior: keep-listed flows use the session, threaded flows raise. Reproduced on this branch:

```python
m._provider = mock            # tests/test_manager_core.py pattern (3 sites)
m.provider is mock            # True — keep-listed flows fine
m.session_for_vm('bprj__p__bprj_c_v')
# ValueError: no provider session registered for provider 'libvirt' (needed by cluster 'c')
```

Those `test_manager_core` tests still pass only because `boxman.manager.Process` is patched there, so the now-threaded `_clone` closure never executes — the divergence is currently masked, not absent.

Options: (a) have `session_for_cluster` fall back to `self._provider` (debug-logged) when nothing is registered for the *primary* type — exact old-behavior parity; or (b) migrate the direct `_provider` pokes to the setter in this PR so "default set ⇒ resolvable" actually holds as an invariant. Nit while here: the annotation says `-> "ProviderSession"` but this fallback can return `None` (nothing registered) — `Optional["ProviderSession"]` matches reality and the call sites' old NoneType-crash parity.

src/boxman/manager.py
```diff
@@ -78,24 +86,147 @@ def load_app_config(self, config: dict[str, Any]) -> None:
...
+
+        Uses the same name construction as the provision/destroy flows
+        (``bprj__<project>__bprj_<cluster>_<vm>``) so call sites that
+        only hold a full VM name (e.g. the post-provision start retry
+        loop) can resolve the owning cluster without parsing names.
+
+        Returns:
+            Mapping of full VM name to cluster name.
+        """
+        prj_name = f'bprj__{self.config["project"]}__bprj'
```
mherkazandjian: **[minor] `session_for_vm`'s documented contract ("names that cannot be derived from the current config fall back to the default session") isn't met when the config itself is underivable:** this line raises `KeyError: 'project'` for configs lacking a `project` key (and `TypeError` for `config=None` — despite the `(self.config or {})` guard two lines down suggesting None-tolerance). Reproduced on this branch with an import-image-style config (`manager.config = boxman_config['providers'][ptype]` has neither `project` nor `clusters`). Unreachable today only because import-image is keep-listed; an early bail (`if not self.config or 'project' not in self.config: return {}`) makes the fallback promise real.

Forward-looking footnote: the comprehension lets colliding full names (`cluster='a', vm='b_c'` vs `cluster='a_b', vm='c'` → same `bprj__p__bprj_a_b_c`) silently resolve to whichever cluster iterates last. Benign now (such configs already collide as libvirt domain names, and all clusters share one session) — worth a comment or guard before Phase 2 lets providers diverge per cluster.

src/boxman/scripts/app.py
```diff
@@ -504,31 +500,45 @@ def main():
...
+            provider_type = primary_provider_type(manager.config)
+
+        # Build a session for every provider declared in the project config
+        # via the registry (import-image keeps its single-provider flow —
+        # the provider type comes from the manifest / CLI flag above).
+        if args.func == BoxmanManager.import_image:
+            provider_types = [provider_type]
+        else:
+            provider_types = (
+                list(manager.config.get('provider', {}).keys()) or [provider_type]
```
mherkazandjian, line 512: **[worth noting] Behavior change for multi-key `provider:` mappings — deserves the same "side effects flagged for review" treatment the vbox fix got.** The old code used only the *first* key and silently ignored the rest; this now eagerly constructs a session for *every* key. Two concrete consequences: (1) `provider: {libvirt: …, virtualbox: …}` previously ran fine as libvirt-only; now it also instantiates `Virtualbox`, whose init wiring reaches toward `vboxmanage` (per the stand-in rationale in `test_provider_registry.py`), on hosts that may not have VirtualBox at all; (2) `provider: {libvirt: …, docker-compose: {}}` previously ran libvirt-only, now exits 2 at startup. Both are defensible — fail-fast beats silently ignoring config, and single-key v1.0 configs are untouched — but under a "(no behavior change)" title this should be listed explicitly in the PR body.

Also note the list is top-level-only: cluster-level `provider:` keys (which `provider_type_for_cluster()` already honors) don't feed it — see the comment on `manager.py:164` for the resulting mid-flow failure.

tests/test_provider_registry.py
```diff
@@ -0,0 +1,162 @@
...
+        manager = self._manager({'provider': {'libvirt': {}}, 'clusters': {}})
+        assert manager.provider_type_for_cluster('nope') == 'libvirt'
+
+    def test_first_registered_session_becomes_default(self):
+        manager = self._manager({'provider': {'libvirt': {}}, 'clusters': {}})
+        session = LibVirtSession(config={"provider": {"libvirt": {}}})
+        manager.register_session('libvirt', session)
+        assert manager.provider is session
+
+    def test_vm_cluster_map_uses_flow_name_construction(self):
```
mherkazandjian, line 148: **[worth addressing] Coverage gaps — the highest-traffic new API is untested.** `_vm_cluster_map()` is pinned here, but `session_for_vm()` — which ~12 threaded call sites (control verbs, up/down, snapshot take/restore, storage compress, provision retry) now route through — has no test for either behavior that matters: (a) map-hit → the cluster's session, (b) not-in-config name → default-session fallback (the PR body's own rationale for removed VMs).

Also untested: commit 2's app.py wiring — the exit-2 CLI contract for `docker-compose`/unknown types (the registry-level error is pinned; the "log cleanly and exit 2" claim isn't), per-key registration, and the vbox `session.manager`-attachment fix the body advertises; plus `register_session` keeping the *first* session as default when a **second** type is registered (the single-registration test can't distinguish first-wins from last-wins). The `session_for_vm` and first-wins cases are two cheap unit tests against a stub manager — worth adding in this PR, since Phase 1's whole deliverable is pinning resolution semantics.

mherkazandjian (commented):
## Phase 1 review — registry & per-cluster dispatch

Exhaustive pass over all three commits: full diff, head-vs-base comparison (`f511fea..59e9f6c`), an audit of every remaining `self.provider`/`cls.provider` site, and the shim edges exercised against the PR head tree (`pytest tests/test_provider_registry.py` → 14 passed here).

### Verdict on the "no behavior change" claim

**Holds for v1.0 single-provider configs on the happy path.** Verified concretely:

- Single `provider:` key → the app.py loop registers exactly one session; `register_session` makes it the default; `provider_type_for_cluster()` returns the primary type for every cluster (no shipped box config has a cluster-level `provider:` key); `session_for_cluster()` / `session_for_vm()` therefore return the *same object* the old `self.provider` returned, at every threaded call site.
- Every `session_for_vm()` input is a config-derived full name (`process_vm_list`, `_select_vm_targets`, `_storage_targets`, `get_connect_info`, `_get_vm_states`-filtered names, `clone_vms` / `_clone_and_configure_new_vms`) — guaranteed `_vm_cluster_map()` hits, so the default-session fallback is never taken for in-config VMs.
- Commit 3 is genuinely 1:1. I checked every substituted site (networks define/remove, clone/destroy, configure+start incl. the correct `cluster_name` parameter, IP/ssh/connect loops, provision retry + eject, up/down, snapshot list/log/take/restore/delete, storage compress, control verbs): same methods, same args, correct loop variables; `_destroy_vm_and_disks` hoists one session for its three calls in unchanged order.
- The 5 `update_provider_config_with_runtime` preambles iterate `_get_sessions().values()` — with one registered session that is equivalent to the old single hasattr-guarded call.
- The libvirt enrichment block in app.py is verbatim (only `provider_type` → `_ptype`), and `primary_provider_type()` preserves the old first-key semantics deterministically (YAML/dict insertion order).
- **The virtualbox latent-bug fix is real**: in the base, `session.manager = manager` / `manager.provider = session` lived inside the libvirt `elif` only, so a vbox session was constructed and dropped. Uniform registration fixes that without touching the libvirt path (`Virtualbox` lacks `update_provider_config_with_runtime`; the `hasattr` guards keep the preambles safe).
- Registry error path confirmed: unknown type → `ValueError` listing supported; `docker-compose` → the friendly Phase-3 message; app.py catches both, `log.error`s, and exits 2 as claimed.

Known deviations — all confined to paths outside "working v1.0 config", but they deserve footnotes given the title says *no* behavior change:

1. Multi-key `provider:` mappings now construct every declared provider eagerly (see inline on `app.py:512`).
2. Configs without a `provider:` section: old code raised `KeyError: 'provider'` at the session-build site; `primary_provider_type()` now defaults to `libvirt` and proceeds (`LibVirtSession.__init__` tolerates the missing block via `.get()`). Strictly friendlier, but a changed failure mode.
3. Code/tests that poke `manager._provider` directly (see inline on `manager.py:209`).

### Findings not anchorable inline (lines outside the diff)

**[worth addressing] Keep-list annotation gaps.** The PR body says every deliberate keep is marked `# Phase 1 (#49)` — there are exactly six markers (import-image :1066, templates :1875, storage pool :2231, removed-VM :2500, collapse :4392, update/diff :5358), but:

- `pxe_boot` (`src/boxman/manager.py:1124`, `session = cls.provider`) is on the body's keep-list yet has **no marker** in code.
- `storage_df` / `storage_trim` / `storage_compact` (+ `storage_optimize`) use `cls.provider.storage` / `cls.provider.provider_config` (manager.py:4435, :4494, :4580) and are **neither marked nor mentioned in the keep-list**, while sibling `storage_compress_snapshots` *was* threaded via `session_for_vm`. Same command family, opposite treatment, zero annotation — exactly the kind of site Phase 3 will trip over. Suggest adding markers (or threading df/trim, which already iterate per-VM).

**[minor] `primary_provider_type()` not adopted inside manager.py.** `_get_vm_states()` (~:3252) and `_find_existing_project_vms()` (~:3181) still hand-roll `list(self.config.get('provider', {}).keys())[0] if 'provider' in self.config else 'libvirt'` — the exact pattern the helper replaces (commit 2's message says "both `list(...)[0]` sites now use `primary_provider_type()`", which is true only of app.py). Worth a same-PR cleanup so resolution logic stays single-sourced.

**[nit]** The body says 13 tests; the file has 14 (`pytest -q` → `14 passed`).

### What's well done

Lazy factories (also what makes the `Virtualbox` monkeypatch test possible), `register_session`'s first-wins default, the two-cluster shared-session AC test, unusually thorough docstrings with phase cross-references, and an honest "side effects flagged for review" section. The threading really is pure call-site substitution — I found no site with changed tuple shapes, arguments, or ordering, and no threaded site left accidentally on the old single session.

src/boxman/scripts/app.py
```diff
@@ -504,31 +500,45 @@ def main():
...
+            if _ptype == 'libvirt':
+                # merge runtime metadata into the provider config from boxman.yml
+                provider_conf_with_runtime = manager.get_provider_config_with_runtime(
+                    boxman_config.get('providers', {}).get(_ptype, {})
+                )
+                # enrich the project config with runtime-aware provider settings
+                # App-level (boxman.yml) settings serve as DEFAULTS;
+                # project-level (conf.yml) settings always take precedence.
+                enriched_config = manager.config.copy()
+                if 'provider' in enriched_config and _ptype in enriched_config['provider']:
```
mherkazandjian: primary_provider_type() now treats a missing or empty provider mapping as libvirt, but this guard discards the app/runtime defaults in exactly that case. For example, a project without a provider section and a boxman.yml URI of qemu+ssh://hv/system creates a session with an empty provider config and silently falls back to local qemu:///system with no sudo, while boxman conf reports the remote URI. Please always create enriched_config["provider"]["libvirt"] from provider_conf_with_runtime, then overlay any project overrides.

src/boxman/manager.py
```diff
@@ -78,24 +86,147 @@ def load_app_config(self, config: dict[str, Any]) -> None:
...
+        Register a live *session* for *provider_type*.
+
+        The first registered session also becomes the default session
+        returned by the legacy :attr:`provider` property.
+
+        Args:
+            provider_type: The provider type name (e.g. ``libvirt``).
+            session: The constructed provider session.
+        """
+        self._get_sessions()[provider_type] = session
```
mherkazandjian, line 156: Keying sessions only by provider type cannot represent the accepted per-cluster session model. Registering a second libvirt session overwrites the first, so both clusters resolve the second session while manager.provider still points at the first. More importantly, ADR-001 requires one Docker Compose project/session per cluster, so two Docker Compose clusters would be forced through the same session state. Please key registrations by cluster (or by provider type plus cluster name) and register sessions per cluster, keeping the default-provider shim separate.

Orski174:
## Final review & merge — Phase 1 (registry & per-cluster dispatch)

Synthesis of the two-round post-review (Copilot + @mherkazandjian / Fable-max). Both reviewers confirmed the core claim — **"no behaviour change" holds for v1.0 single-provider configs on the happy path**, and the commit-3 threading is a pure 1:1 call-site substitution. The findings were edge-path hardening + coverage gaps; all actionable ones are fixed in `df6ee03`, the rest are deferred with rationale below.

### Fixed pre-merge (`df6ee03`)

| # | Finding (reviewer) | Fix |
|---|---|---|
| 1 | `provider` setter not `__new__`-safe — dereferences `self.config` (mher @manager.py:119) | `primary_provider_type(getattr(self, 'config', None))` |
| 2 | `manager.provider is not None` ≠ threaded flows resolve; direct `_provider` pokes diverge (mher @:209) | `session_for_cluster` falls back to `_provider` for the **primary** type when nothing is registered — exact old single-`self.provider` parity; non-primary cluster types still raise the friendly error |
| 3 | `session_for_vm()` annotated non-Optional but can return `None` (Copilot @:210, mher @:209) | retyped `Optional[ProviderSession]` (old NoneType parity documented) |
| 4 | `_vm_cluster_map()` unconditionally reads `config["project"]` → KeyError/TypeError (Copilot @:229, mher @:224) | returns `{}` when config is unset or has no `project`; `session_for_vm` cleanly falls back |
| 5 | `primary_provider_type()` empty-mapping→libvirt discards runtime URI defaults; a no-`provider:` project silently drops to local `qemu:///system` (mher @app.py:525) | libvirt provider block built **unconditionally** from `provider_conf_with_runtime`, project overrides overlaid on top |
| 6 | cluster-level `provider:` honored in manager but app.py builds sessions for top-level keys only → mixed config = mid-flow traceback (mher @:164, @app.py:512) | cluster-level provider types folded into the built set → friendly registry **exit-2** at startup instead of a partial-provision crash |
| 7 | keep-list annotation gaps: `pxe_boot`, `storage_df/trim/compact` unmarked (mher, worth-addressing) | all now carry `# Phase 1 (#49)` markers |
| 8 | `primary_provider_type()` not adopted inside manager.py (5 hand-rolled `list(config['provider'])[0]` sites) (mher, minor) | all 5 single-sourced through the helper (also fixes a latent empty-mapping `IndexError`) |
| 9 | coverage gap — `session_for_vm`, first-wins-with-second-type, `_provider` divergence untested (mher @tests:148, worth-addressing) | **+10 unit tests** (24 total in the file, up from 14) |

### Deferred (with rationale)

- **Per-cluster session keying** (mher @manager.py:144, architectural): keying `_sessions` by provider *type* is correct for Phase 1 — a libvirt host is one connection, so two libvirt clusters *should* share one session. ADR-001's one-compose-project-per-cluster model only bites once `DockerComposeSession` exists, and the `session_for_cluster`/`session_for_vm` seam lets Phase 3 re-key the map internally with **zero call-site churn**. Deferred to **Phase 3 (#51)** with an in-code `.. note::` at the `_sessions` declaration.
- **`VirtualBoxSession` adapter** (Copilot @providers/__init__.py:29): `Virtualbox` doesn't fully implement the `ProviderSession` protocol — but this is a **pre-existing** gap, not introduced here: the base already routed vbox through `self.provider.X`, so this PR's `session_for_cluster().X` calls the identical methods. No regression. The adapter belongs with the provider-protocol work in **Phase 3**.
- **Multi-key top-level `provider:` now eager-instantiates every declared provider** (mher @app.py:512, worth-noting): a behaviour change *only* for non-standard multi-key configs (single-key v1.0 untouched). Fail-fast > silently-ignoring config; documented here rather than reverted.
- **app.py exit-2 CLI contract not unit-tested** (mher, part of the coverage finding): a faithful test needs the `main(argv) -> int` refactor the codebase already flags as a Phase-2.5 follow-up (see `test_cli_smoke.py` docstring). Instead I validated it **live** (below). Tracked, not silently skipped.

### Verification

- **Full non-integration suite: 1129 passed** (1119 + 10 new) on **both** the host and the boxman-provisioned staging VM (`stage01`).
- **Live smoke — happy path** (`snapshot list` against the real staging project via the PR code): session-build loop → enrichment → `session_for_cluster('staging').snapshot_list()` lists the `pristine` snapshot via `virsh -c qemu:///system` — the correct runtime URI, confirming fix #5 end-to-end.
- **Live smoke — fix #6** (synthetic cluster-level `provider: docker-compose`): exits **2** at startup with the friendly Phase-3 message, no mid-flow traceback.

Merging into `feat/docker-compose-provider`. Phase-1 drift retro to follow on #49.

PR #60, opened 2026-07-14
[dc-provider] Phase 2: config schema v2.0 — version key, boxes:, per-cluster provider:

Orski174:
## Phase 2 — config schema v2.0 (#50)

Part of the docker-compose provider epic (#42). Blocked by #49 (merged). Targets `feat/docker-compose-provider`.

### What
Adds versioned configs so a single project can eventually mix libvirt and docker-compose clusters, **without touching v1.0 behaviour**:

- **`version:` key** — absent or `'1.0'` → unchanged; `'2.0'` → normalized; anything else → clean `ConfigError` (exit 2).
- **`boxes:`** as the generic per-cluster key. `BoxmanManager.load_config` normalizes `boxes → vms` for libvirt clusters **once, right after YAML parse**, so all **35** existing `cluster['vms']` consumers are untouched.
- **Per-cluster `provider:`** — already honoured by Phase 1's `provider_type_for_cluster`; now official, validated, and documented. Defaults to the top-level primary provider.

### Compatibility matrix (enforced in `normalize_v2_config`)
| cluster provider | `boxes:` | `vms:` | result |
|---|:---:|:---:|---|
| libvirt | ✓ | – | renamed to `vms:` |
| libvirt | – | ✓ | accepted + deprecation warning |
| libvirt | ✓ | ✓ | `ConfigError` (ambiguous) |
| docker-compose | ✓ | – | kept as-is (Phase 3 consumes it) |
| docker-compose | – | ✓ | `ConfigError` |

A `boxes:` key in a **v1.0** config gets a log-only warning (v1.0 provisioning stays byte-identical).

### Changes
- `manager.py`: `_apply_config_version` + `normalize_v2_config` (+ a v1.0 `boxes:` nudge), called at the end of `load_config`. The `.rendered.yml` dump (raw rendered text) is unchanged.
- `app.py`: `main()` is now a thin wrapper turning `ConfigError` into `log.error` + exit 2 (purely additive — nothing else raises `ConfigError`); body moved verbatim to `_main()`.
- `doc/docker-compose-provider/config-schema.md`: version detection, boxes/vms, per-cluster provider, compat matrix, and the `${VAR}`-safe / bare-`{{ word }}`-unsafe templating caveat for compose values.
- `tests/test_boxman_conf.py`: `TestConfigSchemaV2` (+16).

### Acceptance criteria (#50)
- [x] v1.0 configs: byte-identical behaviour
- [x] v2.0 libvirt-only config behaves exactly like its v1.0 equivalent
- [x] `.rendered.yml` debug dump still written

### Verification
- Full non-integration suite **1145 passed** (1129 + 16 new) on **both** the host and the boxman-provisioned staging VM.
- **Live smoke** against the real staging project: a `boxes:`-keyed v2.0 copy of the staging conf → `snapshot list` resolved the real libvirt domain and `pristine` snapshot via `virsh -c qemu:///system` (proves the normalized v2.0 path drives the identical libvirt resolution as v1.0); `boxman conf` shows the on-disk `boxes:` (fidelity view of the rendered template, by design).

### Known deferral
Structure-aware templating exemption for compose `environment:`/`command:` blocks (bare `{{ word }}` corruption) → **Phase 3**, where those values are actually consumed. Documented in `config-schema.md`; `${VAR}`/`$${VAR}` interpolation is safe today.

### DIFF
--- a/doc/docker-compose-provider/config-schema.md
+++ b/doc/docker-compose-provider/config-schema.md
@@ -21,9 +21,15 @@ of the on-disk schema version.
 | `version:` value | Behaviour |
 |---|---|
 | absent | treated as **v1.0** — returned unchanged |
-| `'1.0'` | **v1.0** — returned unchanged |
-| `'2.0'` | **v2.0** — normalized (see below) |
+| `'1.0'` (or unquoted `1` / `1.0`) | **v1.0** — returned unchanged |
+| `'2.0'` (or unquoted `2` / `2.0`) | **v2.0** — normalized (see below) |
 | anything else | `ConfigError: unsupported config version: '<x>' (supported: '1.0', '2.0')` → CLI logs it and exits 2 |
+
+The value is compared as a string, so unquoted YAML numerics are accepted
+(`version: 2` and `version: 2.0` both select v2.0). **Quoting is still
+recommended** — write `version: '2.0'` — to avoid surprises from YAML's
+numeric coercion (e.g. a future `version: 2.10` would parse as the float
+`2.1`).
 
 ## What v2.0 adds
 
@@ -52,10 +58,40 @@ For each cluster, the effective provider is
 | libvirt | ✓ | ✓ | **`ConfigError`** — ambiguous, declare one |
 | docker-compose | ✓ | – | `boxes:` kept as-is (consumed by the Phase 3 `DockerComposeSession`) |
 | docker-compose | – | ✓ | **`ConfigError`** — docker-compose clusters must use `boxes:` |
+| `virtualbox` | any | any | **`ConfigError`** — not supported under v2.0 yet; use `version: '1.0'` + `vms:` |
+| unknown / typo | any | any | **`ConfigError`** — unknown provider (guards against silent empty clusters) |
+
+Only `libvirt` and `docker-compose` are wired for v2.0. The legacy
+`virtualbox` provider stays on v1.0 syntax for now, and an unknown value
+(e.g. a typo `libvrit`) is rejected rather than left to silently provision
+an empty cluster.
+
+A **per-box `provider:`** key is also rejected (`ConfigError`) — providers
+are a cluster-level concern (ADR-001), so `boxes: { web: { provider: … } }`
+is a config error, not a silently-ignored field.
 
 A v1.0 config that uses `boxes:` gets a one-line warning (nothing reads
-`boxes:` in v1.0, so the cluster would otherwise silently appear empty) —
-provisioning behaviour is unchanged.
+`boxes:` in v1.0, so those boxes are silently ignored — including in a
+partial migration that still has a sibling `vms:`) — provisioning behaviour
+is unchanged.
+
+### Not yet runnable: docker-compose / mixed configs
+
+A v2.0 config with a docker-compose cluster **parses and normalizes**, but
+is **not runnable until Phase 3** (`DockerComposeSession`). Running
+`boxman provision` / `up` on one exits **2** at session build with the
+friendly "docker-compose provider … lands in Phase 3 (#51)" message (the
+registry has no docker-compose session yet), rather than provisioning
+anything. Pure-libvirt v2.0 configs are fully runnable.
+
+### Why `boxman conf` still shows `boxes:`
+
+Normalization (`boxes:` → `vms:`) happens on the in-memory config **after**
+the `.rendered.yml` debug dump is written. `boxman conf` and `.rendered.yml`
+therefore intentionally show the **pre-normalization** shape — a fidelity
+view of what you wrote (`boxes:`), not boxman's internal `vms:` form. This
+is by design; the internal normalization is what the provisioning flows act
+on.
 
 ## Templating caveat for compose values (`environment:`, `command:`)
 
--- a/doc/docker-compose-provider/config-schema.md
+++ b/doc/docker-compose-provider/config-schema.md
@@ -1,0 +1,113 @@
+# Boxman configuration schema — versioning (v1.0 & v2.0)
+
+Boxman project configs (`conf.yml`) carry an optional top-level `version:`
+key. It selects how boxman interprets the rest of the file. **v1.0 is
+supported indefinitely and is unaffected by this document** — everything
+below is additive.
+
+This is the reference for the schema-version handling introduced in
+Phase 2 of the docker-compose provider epic
+([#50](https://github.com/mherkazandjian/boxman/issues/50)). It lives
+alongside [`design.md`](./design.md) (see *Configuration Schema v2.0* and
+*Breaking Changes & Versioning*).
+
+## Version detection
+
+`BoxmanManager.load_config` renders the Jinja template, parses the YAML,
+then calls `_apply_config_version(conf)` before returning — so **every
+downstream consumer sees the internal (v1.0-shaped) config**, regardless
+of the on-disk schema version.
+
+| `version:` value | Behaviour |
+|---|---|
+| absent | treated as **v1.0** — returned unchanged |
+| `'1.0'` | **v1.0** — returned unchanged |
+| `'2.0'` | **v2.0** — normalized (see below) |
+| anything else | `ConfigError: unsupported config version: '<x>' (supported: '1.0', '2.0')` → CLI logs it and exits 2 |
+
+## What v2.0 adds
+
+1. **`boxes:` as the generic per-cluster key.** v1.0 uses `vms:`; v2.0
+   uses `boxes:` (a VM *or* a container is a "box"). For **libvirt**
+   clusters boxman renames `boxes:` → `vms:` at load time, so the entire
+   existing libvirt code path is untouched.
+2. **Per-cluster `provider:`.** Each cluster may declare its own
+   `provider:` (`libvirt` or `docker-compose`). It defaults to the
+   **top-level primary provider** (the first key under the top-level
+   `provider:` mapping). This key is already honoured by the Phase 1
+   provider registry (`provider_type_for_cluster`).
+
+A v2.0 config that uses only libvirt behaves **exactly** like its v1.0
+equivalent.
+
+## Normalization rules (compatibility matrix)
+
+For each cluster, the effective provider is
+`cluster.get('provider') or <top-level primary provider>`. Then:
+
+| cluster provider | `boxes:` | `vms:` | result |
+|---|:---:|:---:|---|
+| libvirt | ✓ | – | `boxes:` renamed to `vms:` |
+| libvirt | – | ✓ | accepted with a **deprecation warning** (prefer `boxes:`) |
+| libvirt | ✓ | ✓ | **`ConfigError`** — ambiguous, declare one |
+| docker-compose | ✓ | – | `boxes:` kept as-is (consumed by the Phase 3 `DockerComposeSession`) |
+| docker-compose | – | ✓ | **`ConfigError`** — docker-compose clusters must use `boxes:` |
+
+A v1.0 config that uses `boxes:` gets a one-line warning (nothing reads
+`boxes:` in v1.0, so the cluster would otherwise silently appear empty) —
+provisioning behaviour is unchanged.
+
+## Templating caveat for compose values (`environment:`, `command:`)
+
+`load_config` pre-processes bare `{{ name }}` tokens into `{name}` task
+placeholders before Jinja render (see the comment block in
+`load_config`). This is a **whole-file** text substitution with no
+awareness of YAML structure, so it also touches docker-compose cluster
+blocks.
+
+Safe / unsafe forms inside a compose `environment:` or `command:` value:
+
+| Form | Matched by preprocessing? | Verdict |
+|---|---|---|
+| `${VAR}`, `$${VAR}` | no (`$`/braces differ) | **safe** — use these for compose-time interpolation |
+| `{{ env("VAR") }}`, `{{ a.b }}`, `{{ x \| y }}` | no (contains non-word chars) | **safe** — rendered by Jinja as usual |
+| `{{ word }}` (a single bare word) | **yes** → rewritten to `{word}` | **avoid** — becomes the literal string `{word}` |
+
+**Rule of thumb:** in docker-compose clusters, use `${VAR}` /
+`$${VAR}` for interpolation and never a bare single-word `{{ word }}`.
+(A structure-aware exemption for compose blocks is deferred to Phase 3,
+where these values are actually consumed.)
+
+## Examples
+
+v1.0 (unchanged, supported indefinitely):
+
+```yaml
+project: my_lab
+provider:
+  libvirt:
+    uri: qemu:///system
+clusters:
+  compute:
+    vms:
+      node01:
+        hostname: node01
+```
+
+v2.0, libvirt-only (behaves identically to the v1.0 above):
+
+```yaml
+version: '2.0'
+project: my_lab
+provider:
+  libvirt:
+    uri: qemu:///system
+clusters:
+  compute:
+    boxes:
+      node01:
+        hostname: node01
+```
+
+v2.0, mixed providers: see
+[`example-conf.yml`](./example-conf.yml).
--- a/src/boxman/manager.py
+++ b/src/boxman/manager.py
@@ -20,7 +20,7 @@ from boxman.config_cache import BoxmanCache
 from boxman.exceptions import ConfigError
 from boxman.image_cache import ImageCache
 from boxman.netlab import ContainerlabManager, shared_bridges
-from boxman.providers import primary_provider_type
+from boxman.providers import PROVIDERS, primary_provider_type
 from boxman.providers.libvirt.commands import VirshCommand
 from boxman.runtime import RuntimeBase, create_runtime
 from boxman.task_runner import TaskRunner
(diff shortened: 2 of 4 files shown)


src/boxman/manager.py
```diff
@@ -433,6 +434,124 @@ def _preserve_jinja_vars(m: re.Match) -> str:
...
+        Returns:
+            The version-normalized configuration.
+
+        Raises:
+            ConfigError: If ``version:`` is set to an unsupported value.
+        """
+        if not conf:
+            return conf
+
+        version = str(conf.get('version', '1.0'))
```
mherkazandjian: **[Medium] Type-coercion papercut, and the one real v1.0 behavior change in this PR.**

`str()` here means unquoted YAML `version: 2.0` works only because `str(2.0) == '2.0'`, while the equally natural `version: 2` → `'2'` and `version: 1` → `'1'` both fall through to the `ConfigError`. So a well-meaning v1.0 user writing `version: 1` gets exit 2 — and note that *before* this PR any top-level `version:` key was silently ignored, so this is the one edge where the "byte-identical v1.0 behavior" claim doesn't hold.

Suggest either normalizing numerics (`if version in ('1', '1.0')` / `('2', '2.0')`) or at least documenting the quoting requirement in `config-schema.md`. There's no test for an unquoted numeric `version:`.

mherkazandjian (commented):
## Review — Phase 2 config schema v2.0

**Verdict:** solid, well-scoped change. I verified the headline claim ("zero v1.0 behavior change") against the code and it **essentially holds** — the v1.0 path returns the same object unchanged (only a log warning is added for `boxes:`), normalization runs after the `.rendered.yml` dump, the entry point (`pyproject.toml`: `boxman = "boxman.scripts.app:main"`) hits the new wrapper, and `ConfigError` is raised nowhere else in `src/`, so the `main()` wrapper is inert for existing flows. **One real exception**: any config carrying a top-level `version:` key with a value other than `'1.0'`/`'2.0'` (including the perfectly reasonable unquoted `version: 1`) was silently ignored before this PR and now hard-fails with exit 2 — see M1.

No blocking issues. Findings ranked:

### Medium
1. **M1 — `version:` type coercion papercut / v1.0 edge regression** (`src/boxman/manager.py:468`). `str(conf.get('version', '1.0'))` makes unquoted YAML `version: 2.0` work only by float-repr accident, while `version: 2` → `'2'` and `version: 1` → `'1'` both raise `ConfigError`. Pre-PR, a stray `version: 1` key was ignored and the config loaded fine; now it exits 2. Suggest accepting `{'1','1.0'}` / `{'2','2.0'}` (or normalizing numerics) and documenting the quoting rule. No test covers unquoted numeric versions.
2. **M2 — unknown / legacy provider silently skips validation** (`src/boxman/manager.py:529`). Only `libvirt` and `docker-compose` branches exist. `primary_provider_type` never returns `None` (defaults to `'libvirt'`), but a cluster with `provider: virtualbox` (a registered provider in `PROVIDERS`) or a typo (`provider: libvrit`) falls through both branches: `boxes:` is left untouched, nothing reads it, and the cluster silently provisions zero VMs — exactly the failure mode `_warn_on_v1_boxes` exists to prevent. Suggest validating against `PROVIDERS` keys and raising `ConfigError` for unknown values (and deciding explicitly what virtualbox+`boxes:` means). No test exercises a non-libvirt/non-dc provider.

### Low
3. **L1 — the app.py change is untested** (`src/boxman/scripts/app.py:297`). All 16 new tests call `_apply_config_version` / `normalize_v2_config` / `load_config` directly; nothing drives `main()` and asserts `SystemExit` code 2 + the logged message. A 5-line test monkeypatching `_main` to raise `ConfigError` would cover the only behavioral change in this file.
4. **L2 — schema now admits configs that crash current commands** (`src/boxman/manager.py:547`). A v2.0 mixed config (libvirt primary + one dc cluster keeping `boxes:`) loads cleanly, but ~15 downstream sites do unguarded `cluster['vms']` over *all* clusters (e.g. `vm_clone_tasks`, manager.py:2404) → raw `KeyError: 'vms'` traceback on `boxman provision`/`up`. Understood that Phase 3 filters clusters per provider; until then a one-line note in `config-schema.md` ("mixed configs are not yet runnable") or a temporary guard would avoid confusing tracebacks. Also: `create_session('docker-compose', …)` raising `NotImplementedError` (not `ConfigError`) means a dc-primary config escapes the clean exit-2 path.

### Nits
5. **N1 — `normalize_v2_config` is not idempotent** (`src/boxman/manager.py:540`). `cluster.pop('boxes')` mutates in place while `conf['version']` stays `'2.0'`; a second pass over the same dict would emit a spurious "legacy 'vms:'" deprecation warning. Only theoretical today (`load_config` re-parses per call) — worth a docstring warning at minimum.
6. **N2 — misleading dc error for the both-keys case** (`src/boxman/manager.py:550`). A dc cluster declaring both `boxes:` and `vms:` gets "must use 'boxes:', not 'vms:'" — but it *does* use `boxes:`. The libvirt branch has a dedicated ambiguity message; the dc branch could reword or mirror it. The raise itself is correct.
7. **N3 — loose test assertion** (`tests/test_boxman_conf.py:907`). `"legacy 'vms:'" in r.message or "vms:" in r.message` — the second disjunct subsumes the first and matches nearly any warning mentioning vms. Missing coverage worth adding while here: mixed libvirt+dc config end-to-end, dc cluster with both keys, unquoted `version: 2`, unknown per-cluster provider.
8. **N4 — doc gaps** (`doc/docker-compose-provider/config-schema.md:26`). The doc should state (a) the quoting caveat from M1, and (b) that `.rendered.yml` / `boxman conf` intentionally show the pre-normalization `boxes:` shape (the PR body says "by design" but the doc doesn't).

### Verified claims (for the record)
- v1.0 path byte-identical: **confirmed** (modulo the M1 `version:`-key edge).
- `.rendered.yml` still written, not re-read anywhere in `src/`: **confirmed** (only referenced in `cli_parser.py` help text).
- Provider resolution `cluster.get('provider') or primary_provider_type(conf)` matches Phase 1's `provider_type_for_cluster`: **confirmed**.
- Entry point still `main`: **confirmed**.
- `captured_logs` fixture pre-exists in `tests/conftest.py`: **confirmed**.
- `example-conf.yml` link target exists: **confirmed**.

src/boxman/manager.py
```diff
@@ -433,6 +434,124 @@ def _preserve_jinja_vars(m: re.Match) -> str:
...
+            The normalized configuration (same object).
+
+        Raises:
+            ConfigError: On an ambiguous or provider-incompatible cluster.
+        """
+        for cluster_name, cluster in (conf.get('clusters') or {}).items():
+            if not isinstance(cluster, dict):
+                continue
+
+            provider = cluster.get('provider') or primary_provider_type(conf)
```
mherkazandjian, line 554: **[Medium] Unknown/legacy providers silently skip validation.**

Only `'libvirt'` and `'docker-compose'` are handled below. `primary_provider_type` never returns `None` (it defaults to `'libvirt'`), but two real cases fall through both branches with no error and no warning:

- `provider: virtualbox` — a *registered* provider in `PROVIDERS`; a `boxes:` key is left untouched, nothing reads it, cluster silently provisions zero VMs.
- a typo, e.g. `provider: libvrit` — same silent-empty outcome, and the typo is only caught much later (if at all) by `create_session`.

That silent-empty-cluster failure mode is exactly what `_warn_on_v1_boxes` exists to prevent. Suggest a final `else: raise ConfigError(f"cluster '{cluster_name}': unknown provider '{provider}' …")` validated against `PROVIDERS` keys (and an explicit decision for virtualbox+`boxes:`). No test covers a non-libvirt/non-dc provider.

src/boxman/manager.py
```diff
@@ -433,6 +434,124 @@ def _preserve_jinja_vars(m: re.Match) -> str:
...
+            has_vms = 'vms' in cluster
+
+            if provider == 'libvirt':
+                if has_boxes and has_vms:
+                    raise ConfigError(
+                        f"cluster '{cluster_name}' declares both 'boxes:' "
+                        f"and 'vms:' — use one (prefer 'boxes:' in v2.0)."
+                    )
+                if has_boxes:
+                    cluster['vms'] = cluster.pop('boxes')
```
mherkazandjian, line 570: **[Nit] Not idempotent.** This `pop` mutates the parsed dict in place while `conf['version']` stays `'2.0'`, so normalizing the same object twice would take the `has_vms` branch on the second pass and emit a spurious "legacy 'vms:'" deprecation warning. Only theoretical today — `load_config` re-parses fresh each call and nothing else invokes `normalize_v2_config` — but since the method is public, a docstring warning (or resetting `conf['version']` post-normalization) would be cheap insurance.

src/boxman/manager.py
```diff
@@ -433,6 +434,124 @@ def _preserve_jinja_vars(m: re.Match) -> str:
...
+                elif has_vms:
+                    self.logger.warning(
+                        f"cluster '{cluster_name}' uses the legacy 'vms:' "
+                        f"key under version: '2.0' — 'boxes:' is the "
+                        f"preferred generic key. 'vms:' remains accepted."
+                    )
+            elif provider == 'docker-compose':
```
mherkazandjian, line 577: **[Low] The schema now admits configs that current commands crash on.**

A v2.0 mixed config (libvirt primary + one dc cluster) loads cleanly with `boxes:` kept here, but ~15 downstream sites do unguarded `cluster['vms']` over *all* clusters — e.g. `vm_clone_tasks` (manager.py:2404), and similar loops at 2495, 2672, 3119… — so `boxman provision`/`up` on such a config dies with a raw `KeyError: 'vms'` traceback rather than a clean message. I get that Phase 3 filters clusters per provider; until then, a one-line "mixed configs are not yet runnable" note in `config-schema.md`, or a temporary guard, would avoid confusing tracebacks. Relatedly, a dc-*primary* config hits `NotImplementedError` from `create_session`, which bypasses the new clean `ConfigError` → exit-2 path.

src/boxman/manager.py
```diff
@@ -433,6 +434,124 @@ def _preserve_jinja_vars(m: re.Match) -> str:
...
+                if has_vms:
+                    raise ConfigError(
+                        f"cluster '{cluster_name}' is a docker-compose "
```
mherkazandjian: **[Nit] Misleading message when a dc cluster declares *both* keys.** This branch fires on `has_vms` alone, so a cluster with both `boxes:` and `vms:` is told it "must use 'boxes:', not 'vms:'" — but it *does* use `boxes:`. The libvirt branch has a dedicated ambiguity message for the both-keys case; consider mirroring it, or rewording to "declares 'vms:'; docker-compose clusters only support 'boxes:'". The raise itself is correct.

src/boxman/scripts/app.py
```diff
@@ -282,6 +283,23 @@ def load_boxman_config(path: str) -> dict:
...
+
+    Thin wrapper that translates config-schema errors
+    (:class:`~boxman.exceptions.ConfigError`, e.g. an unsupported
+    ``version:`` or an invalid v2.0 cluster) into a clean ``log.error`` +
+    exit 2, instead of surfacing a traceback. All other flow (including
+    the ``sys.exit()`` calls throughout) lives in :func:`_main`.
+    """
+    try:
+        _main()
+    except ConfigError as exc:
```
mherkazandjian, line 297: **[Low] This is the only behavioral change in app.py and it has no test.** All 16 new tests call `_apply_config_version`/`normalize_v2_config`/`load_config` directly; nothing drives `main()` and asserts `SystemExit` with code 2 plus the logged message. A small test that monkeypatches `_main` to raise `ConfigError('boom')` and asserts `pytest.raises(SystemExit)` → `.value.code == 2` (+ `captured_logs`) would lock the wrapper in.

(Verified for the record: `pyproject.toml` still points the console script at `main`, and `ConfigError` is raised nowhere else in `src/`, so the wrapper is inert for all pre-existing flows — the "purely additive" claim holds.)

tests/test_boxman_conf.py
```diff
@@ -823,3 +825,186 @@ def _failing_run(cmd, *_a, **_kw):
...
+        assert out2['clusters'] == out1['clusters']
+
+    def test_v2_libvirt_vms_accepted_with_warning(self, captured_logs):
+        cfg = self._cfg(version='2.0', top_provider='libvirt',
+                        cluster={'vms': {'n': {}}})
+        with captured_logs.at_level(logging.WARNING, logger='boxman'):
+            out = self._mgr()._apply_config_version(cfg)
+        assert out['clusters']['c']['vms'] == {'n': {}}
+        assert any(
+            "legacy 'vms:'" in r.message or "vms:" in r.message
```
mherkazandjian: **[Nit] Loose assertion:** `"legacy 'vms:'" in r.message or "vms:" in r.message` — the second disjunct subsumes the first and matches nearly any warning that mentions vms; drop the `or "vms:"` clause. While in this file, coverage worth adding: (1) a mixed libvirt+dc config in one `clusters:` map, (2) a dc cluster declaring *both* `boxes:` and `vms:`, (3) unquoted `version: 2` / `version: 1` through `load_config`, (4) a cluster with `provider: virtualbox` or a typo (currently silently unvalidated), (5) an idempotency/double-normalize check.

doc/docker-compose-provider/config-schema.md
```diff
@@ -0,0 +1,113 @@
...
+then calls `_apply_config_version(conf)` before returning — so **every
+downstream consumer sees the internal (v1.0-shaped) config**, regardless
+of the on-disk schema version.
+
+| `version:` value | Behaviour |
+|---|---|
+| absent | treated as **v1.0** — returned unchanged |
+| `'1.0'` | **v1.0** — returned unchanged |
+| `'2.0'` | **v2.0** — normalized (see below) |
+| anything else | `ConfigError: unsupported config version: '<x>' (supported: '1.0', '2.0')` → CLI logs it and exits 2 |
```
mherkazandjian, line 26: **[Nit] Two doc gaps:** (1) the version-detection table should warn that the value is string-compared after `str()` coercion — unquoted `2.0` happens to work, but `version: 2` or `version: 1` is rejected, so recommend quoting (`version: '2.0'`). (2) The PR body says the `.rendered.yml` dump / `boxman conf` intentionally show the pre-normalization `boxes:` shape ("fidelity view"), but the doc never states that; one sentence here would preempt "why does boxman conf still show boxes:?" confusion.

src/boxman/manager.py
mherkazandjian, line 570: **[Medium] The accepted per-cluster-provider invariant is not enforced here.** ADR-001 says a `provider:` inside an individual box is a config error and explicitly assigns that validation to Phase 2 (#50), but this normalization accepts it unchanged. For example, `boxes: {web: {provider: docker-compose}}` under a libvirt cluster becomes `vms: {web: {provider: docker-compose}}`; `provider_type_for_cluster()` still routes the box through libvirt, so the requested provider is silently ignored. Please inspect both the `boxes:` mapping and the accepted legacy `vms:` mapping and raise `ConfigError` when any box declares `provider:`.

src/boxman/scripts/app.py
```diff
@@ -282,6 +283,23 @@ def load_boxman_config(path: str) -> dict:
...
+        log.error(str(exc))
```
mherkazandjian, line 303: **[Medium] This error can be silently filtered for `snapshot log`.** That command sets the `boxman` logger to `CRITICAL + 1` before constructing `BoxmanManager` (lines 371–372). If config loading then raises `ConfigError`, control reaches this handler with that level still active, so `log.error(...)` emits nothing and the CLI exits 2 with an empty stdout. I reproduced exactly `{exit_code: 2, stdout: ""}`. Please restore the logger before reporting the error, ideally with a `finally` around the snapshot-log suppression, or report via a channel that cannot be suppressed.

src/boxman/manager.py
```diff
@@ -433,6 +434,124 @@ def _preserve_jinja_vars(m: re.Match) -> str:
...
+        """
+        Warn when a v1.0 config uses the v2.0-only ``boxes:`` key.
+
+        In v1.0 nothing reads ``boxes:``, so such a cluster would silently
+        appear to have no VMs. This is a log-only nudge toward
+        ``version: '2.0'`` — it does not change v1.0 provisioning
+        behaviour.
+        """
+        for cluster_name, cluster in (conf.get('clusters') or {}).items():
+            if isinstance(cluster, dict) and 'boxes' in cluster and 'vms' not in cluster:
```
mherkazandjian: **[Medium] This skips the migration warning when `vms:` is also present.** A plausible partial migration with `vms: {old: {}}` plus `boxes: {new: {}}` emits zero warnings; existing flows then provision only `old` and silently ignore `new`. That contradicts the documented promise that v1 `boxes:` usage gets a warning. Please drop the `and "vms" not in cluster` condition and word the warning generically so every v1 `boxes:` use is surfaced.

Orski174:
## Final review & merge — Phase 2 (config schema v2.0)

Synthesis of the two-round post-review (Copilot + @mherkazandjian / Fable-max). **No blocking issues**; the v1.0-byte-identical claim holds (verified by mher) except for the one edge M1 fixes. Every actionable finding is addressed in `53a307c`; the one stale finding (L2) is confirmed no-repro.

### Fixed pre-merge (`53a307c`)

| # | Finding | Fix |
|---|---|---|
| **M1** | `str(version)` rejected unquoted `version: 1` / `2` — the one real v1.0 regression (a stray `version:` was ignored pre-PR). | Accept `{1,'1','1.0'}`/`{2,'2','2.0'}`; string-compare after coercion; quoting doc'd. |
| **M2/C2** | Non-libvirt/non-dc provider (virtualbox, or a typo) silently left `boxes:` untouched → empty cluster. | Reject: virtualbox → "use `version:'1.0'`+`vms:`"; unknown → "unknown provider". |
| **M3** | Per-box `provider:` accepted though ADR-001 assigns its rejection to Phase 2. | `ConfigError` on any box declaring `provider:` (both `boxes:` and legacy `vms:` mappings). |
| **M4** | `snapshot log` raises the logger to `CRITICAL+1`, so the wrapper's `log.error` was swallowed → exit 2 with empty output. | Restore the boxman logger level before reporting. |
| **M5** | `_warn_on_v1_boxes` skipped the warning when `vms:` was also present → partial migration silently dropped the `boxes:`. | Warn on any v1 `boxes:` usage; generic wording. |
| **C1** | Non-mapping YAML root → `AttributeError` past the clean handler. | `if not isinstance(conf, dict): return conf`. |
| **N1/N2/N3/N4** | Idempotency contract undocumented; misleading dc both-keys message; loose test assertion; doc gaps (quoting + "conf shows pre-norm `boxes:` by design"). | All addressed. |
| **L1** | The `main()` wrapper (only app.py behavior change) was untested. | Added exit-2 test + the M4 suppression case. |

### Verified, not fixed
- **L2 (mixed-config `KeyError`)** — **stale / no CLI repro.** Phase 1's cluster-fold (app.py:536-539) appends the dc cluster's type → `create_session('docker-compose')` → `NotImplementedError` → **exit 2 at session build**, before provisioning ever reaches `cluster['vms']`. Confirmed by driving the exact `provider_types` computation on a mixed config. Added a `config-schema.md` note that docker-compose/mixed configs parse but aren't runnable until Phase 3 — no throwaway guard.

### Coverage added
+15 tests (72 in the file): numeric versions (unit + `load_config`), non-dict root, virtualbox/unknown provider, per-box provider (both mappings), partial-migration warning, dc both-keys message, mixed-config parse, double-normalize idempotency, and the two `main()`-wrapper cases.

### Verification
Full non-integration suite **1160 passed** (1145 + 15) on **both** the host and the staging VM. No live host libvirt this phase (pure config change; unit tests cover normalization), per the epic's host-vs-VM testing split.

Merging into `feat/docker-compose-provider`. Phase-2 drift retro to follow on #50.

PR #61, opened 2026-07-15
feat(docker-compose): DockerComposeSession core lifecycle (#51)

Orski174:
## Phase 3 — DockerComposeSession core lifecycle (#51)

Part of the docker-compose provider epic (#42). Phases 0–2 (design, provider
registry, config schema v2.0) are already on `feat/docker-compose-provider`.
Before this PR, a v2.0 config could *declare* a `provider: docker-compose`
cluster but running it exited 2 (no session existed). **Phase 3 makes
`boxman provision / up / down / deprovision / destroy` actually work for a
docker-compose cluster** — default docker **bridge** networks only.

Out of scope (later phases): macvlan/shared-bridge L2 to VMs (#52, Phase 4),
structured volumes (Phase 5), ssh/exec/inventory (Phase 6), snapshots via
`docker commit` (Phase 7).

### Architecture — coarse per-cluster seam
boxman's lifecycle flows are VM-centric (per-VM loops over `cluster['vms']`,
each in its own `multiprocessing.Process`). docker-compose is cluster-centric
(one file + one `docker compose up --wait` per cluster, ADR-001/D1/D5). Rather
than reshape the libvirt flows, the manager **dispatches a whole dc cluster to
session-level coarse methods**, and the existing per-VM/network helpers **skip**
dc clusters. Libvirt flows stay byte-identical.

### New package `src/boxman/providers/docker_compose/`
- **`ComposeGenerator`** — translate `boxes:` → compose `services:`
  (`image`/`command`/`environment`/`ports`/`depends_on`/`restart`/`healthcheck`);
  resolve `build.context` absolute vs the conf.yml dir (D4); cluster-internal
  `networks:` → top-level compose networks (`driver: bridge` + ipam subnet);
  deep-merge per-cluster & per-box `compose_extra:` verbatim (D7); **warn+skip**
  out-of-phase box features (`volumes:` → Phase 5; shared/macvlan network refs →
  Phase 4). Writes `<workdir>/docker-compose.yml` (D5), an inspectable,
  hand-runnable fidelity artifact.
- **`ComposeRunner`** — stateless; shells out to host `docker compose` (mirrors
  `ContainerlabManager` / the docker-compose *runtime*'s host-side calls).
  `preflight()`, `up -d --wait --wait-timeout <readiness_timeout|120>` (D1) →
  `ProvisionError` on failure, `down` / `down --volumes` / `stop` / `start` / `ps`.
- **`DockerComposeSession`** — satisfies the `ProviderSession` protocol
  (`provider_config`/`uri`/`use_sudo`/`update_provider_config*`); coarse
  `up_cluster` / `stop_cluster` / `start_cluster` / `down_cluster` /
  `destroy_cluster`. The per-VM/libvirt-shaped protocol methods
  (`clone_vm`, `start_vm`, `define_network`, `snapshot_*`, …) raise a clear
  "docker-compose is cluster-scoped; use the cluster lifecycle" error — they are
  never reached because the manager takes the coarse path.

### Registry + guardrail
- Registry now returns a real `DockerComposeSession` (Phase-1 stub removed).
- **`runtime: local` guardrail** in `app.py`: a dc-provider cluster with a
  non-local runtime fails fast (exit 2). The `docker-compose` *runtime*
  (libvirt-in-a-container) is a different axis — see `config-schema.md`.
  Defense-in-depth: `up_cluster` re-checks via `self.manager.runtime`.

### Manager integration (libvirt byte-identical)
- New `_vm_clusters` view + `_compose_clusters()` / `_is_compose_cluster()`.
  All per-VM/network/ssh helpers iterate `_vm_clusters` (dc clusters carry
  `boxes:`, not `vms:`, so they're skipped structurally — no `KeyError`).
- Coarse `provision_/stop_/start_/deprovision_/destroy_compose_clusters()`
  helpers wired into provision/up/down/deprovision/destroy (ordering mirrors
  the netlab hooks — dc up after libvirt VMs, torn down before). Cache
  registration is provider-agnostic (unchanged).
- Cosmetic: dc-only `down` no longer logs the libvirt "saving all VMs to disk"
  line when there are no VMs.

### Verification
- **Host unit suite: 1209 passed** (+49 new in `tests/test_docker_compose_provider.py`:
  generator translation, runner command strings, session protocol + coarse
  dispatch, the guardrail, and manager partitioning — all mocked, no live
  docker/libvirt). Stale Phase-1 registry stub test updated.
- **Live nested dc-only e2e inside stage01** (first live-container phase; never
  on the host hypervisor): `boxman provision` a 2-service bridge-network cluster
  (redis + nginx, both healthchecked) → `docker compose up -d --wait` to
  **both healthy**; `up` idempotent (no recreate); `down`→`stop` then `up`→healthy;
  `destroy` removes containers + network + the generated compose file. All green.

### DIFF
--- a/src/boxman/manager.py
+++ b/src/boxman/manager.py
@@ -233,6 +233,36 @@ class BoxmanManager:
             return self.provider
         return self.session_for_cluster(cluster_name)
 
+    def _is_compose_cluster(self, cluster_name: str) -> bool:
+        """True if *cluster_name* is provisioned by the docker-compose provider."""
+        return self.provider_type_for_cluster(cluster_name) == 'docker-compose'
+
+    @property
+    def _vm_clusters(self) -> dict[str, Any]:
+        """
+        Clusters handled by a **per-VM** provider (libvirt).
+
+        The manager's per-VM / per-network lifecycle helpers iterate this
+        instead of ``config['clusters']`` so docker-compose clusters (which
+        carry ``boxes:``, not ``vms:``, and are provisioned coarsely via the
+        ``*_compose_clusters`` helpers) are never fed into the per-VM loops.
+        For a libvirt-only project this is exactly ``config['clusters']``, so
+        behaviour is unchanged.
+        """
+        return {
+            name: cluster
+            for name, cluster in (self.config.get('clusters') or {}).items()
+            if not self._is_compose_cluster(name)
+        }
+
+    def _compose_clusters(self) -> dict[str, Any]:
+        """docker-compose clusters, in config order (empty for libvirt-only projects)."""
+        return {
+            name: cluster
+            for name, cluster in (self.config.get('clusters') or {}).items()
+            if self._is_compose_cluster(name)
+        }
+
     def _vm_cluster_map(self) -> dict[str, str]:
         """
         Map every full VM name of this project to its cluster name.
@@ -2370,6 +2400,38 @@ class BoxmanManager:
         netlab.render_topology(source_root=source_root)
         netlab.ensure_up()
 
+    # --- docker-compose clusters: coarse per-cluster lifecycle ------------
+    # docker-compose is cluster-scoped (one `docker compose up --wait` per
+    # cluster, ADR-001/D1). Rather than the per-VM libvirt loops, the manager
+    # dispatches a whole dc cluster to its session's coarse methods. These
+    # helpers are no-ops for libvirt-only projects (``_compose_clusters()``
+    # is empty).
+
+    def provision_compose_clusters(self) -> None:
+        """``docker compose up --wait`` every docker-compose cluster."""
+        for cluster_name, cluster in self._compose_clusters().items():
+            self.session_for_cluster(cluster_name).up_cluster(cluster_name, cluster)
+
+    def stop_compose_clusters(self) -> None:
+        """``docker compose stop`` every docker-compose cluster (boxman down)."""
+        for cluster_name, cluster in self._compose_clusters().items():
+            self.session_for_cluster(cluster_name).stop_cluster(cluster_name, cluster)
+
+    def start_compose_clusters(self) -> None:
+        """``docker compose start`` every docker-compose cluster."""
+        for cluster_name, cluster in self._compose_clusters().items():
+            self.session_for_cluster(cluster_name).start_cluster(cluster_name, cluster)
+
+    def deprovision_compose_clusters(self) -> None:
+        """``docker compose down`` every docker-compose cluster (keep volumes)."""
+        for cluster_name, cluster in self._compose_clusters().items():
+            self.session_for_cluster(cluster_name).down_cluster(cluster_name, cluster)
+
+    def destroy_compose_clusters(self) -> None:
+        """``docker compose down --volumes`` every docker-compose cluster."""
+        for cluster_name, cluster in self._compose_clusters().items():
+            self.session_for_cluster(cluster_name).destroy_cluster(cluster_name, cluster)
+
     def define_networks(self) -> None:
         """
         Define the networks specified in the cluster configuration (sequential).
@@ -2380,7 +2442,7 @@ class BoxmanManager:
         otherwise two concurrent processes can both select the same bridge index
         and the second net-define fails with "bridge name already in use".
         """
-        for cluster_name, cluster in self.config['clusters'].items():
+        for cluster_name, cluster in self._vm_clusters.items():
             for network_name, network_info in cluster.get('networks', {}).items():
                 _network_name = self.full_network_name(
                     project_config=self.config,
@@ -2417,7 +2479,7 @@ class BoxmanManager:
 
         processes = [
             Process(target=_destroy, args=(cluster_name, cluster, network_name, network_info))
-            for cluster_name, cluster in self.config['clusters'].items()
+            for cluster_name, cluster in self._vm_clusters.items()
             for network_name, network_info in cluster.get('networks', {}).items()
         ]
         [p.start() for p in processes]
@@ -2461,7 +2523,7 @@ class BoxmanManager:
         """
         def vm_clone_tasks():
             prj_name = f'bprj__{self.config["project"]}__bprj'
-            for cluster_name, cluster in self.config['clusters'].items():
+            for cluster_name, cluster in self._vm_clusters.items():
                 for vm_name, vm_info in cluster['vms'].items():
                     vm_info = vm_info.copy()
                     new_vm_name = f"{prj_name}_{cluster_name}_{vm_name}"
@@ -2472,7 +2534,7 @@ class BoxmanManager:
         # directory; doing it here sequentially prevents the race condition
         # where N parallel processes all try to define the same pool at once.
         seen_workdirs: set = set()
-        for cluster_name, cluster in self.config['clusters'].items():
+        for cluster_name, cluster in self._vm_clusters.items():
             workdir = os.path.abspath(os.path.expanduser(cluster['workdir']))
             if workdir not in seen_workdirs:
                 seen_workdirs.add(workdir)
@@ -2552,7 +2614,7 @@ class BoxmanManager:
         """
         prj_name = f'bprj__{self.config["project"]}__bprj'
         def vm_destroy_tasks():
-            for cluster_name, cluster in self.config['clusters'].items():
+            for cluster_name, cluster in self._vm_clusters.items():
                 for vm_name in cluster['vms'].keys():
                     full_vm_name = f"{prj_name}_{cluster_name}_{vm_name}"
                     yield full_vm_name, cluster_name, vm_name
@@ -2729,7 +2791,7 @@ class BoxmanManager:
                 target=self._configure_and_start_vm,
                 args=(cluster_name, cluster, vm_name, vm_info)
             )
-            for cluster_name, cluster in self.config['clusters'].items()
+            for cluster_name, cluster in self._vm_clusters.items()
             for vm_name, vm_info in cluster['vms'].items()
         ]
         [p.start() for p in processes]
@@ -2740,7 +2802,7 @@ class BoxmanManager:
     def configure_network_interfaces(self) -> None:
         """Configure network interfaces for all VMs (sequential)."""
         prj_name = f'bprj__{self.config["project"]}__bprj'
-        for cluster_name, cluster in self.config['clusters'].items():
+        for cluster_name, cluster in self._vm_clusters.items():
             for vm_name, vm_info in cluster['vms'].items():
                 full_vm_name = f"{prj_name}_{cluster_name}_{vm_name}"
                 vm_info = vm_info.copy()
@@ -2765,7 +2827,7 @@ class BoxmanManager:
     def configure_disks(self) -> None:
         """Configure disks for all VMs (sequential)."""
         prj_name = f'bprj__{self.config["project"]}__bprj'
-        for cluster_name, cluster in self.config['clusters'].items():
+        for cluster_name, cluster in self._vm_clusters.items():
             workdir = cluster.get('workdir', '.')
             for vm_name, vm_info in cluster['vms'].items():
                 full_vm_name = f"{prj_name}_{cluster_name}_{vm_name}"
@@ -2786,7 +2848,7 @@ class BoxmanManager:
     def configure_cpu_mem(self) -> None:
         """Configure CPU and memory for all VMs (sequential)."""
         prj_name = f'bprj__{self.config["project"]}__bprj'
-        for cluster_name, cluster in self.config['clusters'].items():
+        for cluster_name, cluster in self._vm_clusters.items():
             for vm_name, vm_info in cluster['vms'].items():
                 full_vm_name = f"{prj_name}_{cluster_name}_{vm_name}"
                 self.logger.info(
@@ -2808,7 +2870,7 @@ class BoxmanManager:
     def start_vms(self) -> None:
         """Start all VMs (sequential)."""
         prj_name = f'bprj__{self.config["project"]}__bprj'
-        for cluster_name, cluster in self.config['clusters'].items():
+        for cluster_name, cluster in self._vm_clusters.items():
             for vm_name, vm_info in cluster['vms'].items():
                 full_vm_name = f"{prj_name}_{cluster_name}_{vm_name}"
                 self.logger.info(f"starting vm {full_vm_name}")
@@ -2831,7 +2893,7 @@ class BoxmanManager:
         prj_name = f'bprj__{self.config["project"]}__bprj'
         vm_names = [
             f"{prj_name}_{cluster_name}_{vm_name}"
-            for cluster_name, cluster in self.config['clusters'].items()
+            for cluster_name, cluster in self._vm_clusters.items()
             for vm_name in cluster['vms']
         ]
 
@@ -2865,7 +2927,7 @@ class BoxmanManager:
         ws_path = self.config.get('workspace', {}).get('path', '')
 
         prj_name = f'bprj__{self.config["project"]}__bprj'
-        for cluster_name, cluster in self.config['clusters'].items():
+        for cluster_name, cluster in self._vm_clusters.items():
             self.logger.info(f"cluster: {cluster_name}")
             self.logger.info("-" * 60)
 
@@ -2972,7 +3034,7 @@ class BoxmanManager:
         # blocks — opening 'w' once per cluster would have the later
         # iteration truncate the earlier one's entries.
         groups: dict[str, list[tuple[str, dict, str]]] = {}
-        for cluster_name, cluster in self.config['clusters'].items():
+        for cluster_name, cluster in self._vm_clusters.items():
             base_path = ws_path or cluster.get('workdir', '~')
             ssh_config = os.path.expanduser(os.path.join(
                 base_path,
@@ -3044,7 +3106,7 @@ class BoxmanManager:
         success = True
         ws_path = self.config.get('workspace', {}).get('path', '')
 
-        for _, cluster in self.config['clusters'].items():
+        for _, cluster in self._vm_clusters.items():
             base_path = os.path.expanduser(ws_path or cluster['workdir'])
             admin_key_name = cluster.get('admin_key_name', 'id_ed25519_boxman')
 
@@ -3155,7 +3217,7 @@ class BoxmanManager:
         ws_path = self.config.get('workspace', {}).get('path', '')
 
         prj_name = f'bprj__{self.config["project"]}__bprj'
-        for cluster_name, cluster in self.config['clusters'].items():
+        for cluster_name, cluster in self._vm_clusters.items():
             base_path = os.path.expanduser(ws_path or cluster['workdir'])
             admin_key_name = cluster.get('admin_key_name', 'id_ed25519_boxman')
             admin_pub_key = os.path.join(base_path, f"{admin_key_name}.pub")
@@ -3343,7 +3405,7 @@ class BoxmanManager:
         """
         # write global authorized keys so they can be consumed by container
         # entrypoints or cloud-init scripts
-        for _, cluster in self.config['clusters'].items():
+        for _, cluster in self._vm_clusters.items():
             workdir = os.path.expanduser(cluster['workdir'])
             global_keys_path = os.path.join(workdir, 'global_authorized_keys')
             self.write_global_authorized_keys_file(global_keys_path)
@@ -3631,7 +3693,7 @@ class BoxmanManager:
         # backing chain issues and tray-lock errors on subsequent snapshots.
         cls.logger.info("ejecting cdrom (seed.iso) from all VMs post-provisioning...")
         prj_name = f'bprj__{config["project"]}__bprj'
-        for _cluster_name, _cluster in config['clusters'].items():
+        for _cluster_name, _cluster in cls._vm_clusters.items():
             for _vm_name, _ in _cluster['vms'].items():
                 _full_vm_name = f"{prj_name}_{_cluster_name}_{_vm_name}"
                 cls.session_for_cluster(_cluster_name).eject_cdrom(_full_vm_name)
@@ -3641,6 +3703,10 @@ class BoxmanManager:
 
         # display connection information (after ssh setup so connections are ready)
         cls.connect_info()
+
+        # bring up docker-compose clusters (no-op for libvirt-only projects);
+        # after libvirt VMs, mirroring the netlab hook's "extra infra last" order
+        cls.provision_compose_clusters()
 
         # render and deploy the containerlab topology (no-op if not configured)
         cls.deploy_netlab()
@@ -3664,6 +3730,13 @@ class BoxmanManager:
         expected_vms = cls._get_project_vm_names()
 
         if not expected_vms:
+            # No libvirt VMs. A docker-compose-only project still has work to
+            # do: (idempotently) bring up its compose clusters. `up_cluster`
+            # is `docker compose up -d --wait`, which starts stopped containers
+            # and no-ops when everything is already running.
+            if cls._compose_clusters():
+                cls.provision_compose_clusters()
+                return
             cls.logger.error("no VMs defined in configuration")
             return
 
@@ -3710,6 +3783,9 @@ class BoxmanManager:
             # even though the VMs stayed up.
             cls.ensure_shared_bridges()
             cls.ensure_netlab_up()
+            # Reconcile docker-compose clusters too: a host reboot or manual
+            # `docker compose stop` may have left them down (idempotent).
+            cls.provision_compose_clusters()
             cls.connect_info()
             # Re-write SSH config in case IPs changed (DHCP renewals after
             # a host reboot, manual virsh net cycle, etc.) or in case the
@@ -3793,6 +3869,9 @@ class BoxmanManager:
         # shared bridges have live endpoints on both sides.
         cls.ensure_netlab_up()
 
+        # Bring up docker-compose clusters after the VMs are up (idempotent).
+        cls.provision_compose_clusters()
+
         # Display connection information
         cls.connect_info()
 
@@ -3820,14 +3899,15 @@ class BoxmanManager:
 
         vm_list = cls.process_vm_list(cli_args)
 
-        if not vm_list:
+        if not vm_list and not cls._compose_clusters():
             cls.logger.info("no VMs found in configuration")
             return
 
         use_suspend = getattr(cli_args, 'suspend', False)
 
         if use_suspend:
-            cls.logger.info("suspending all VMs (--suspend)...")
+            if vm_list:
+                cls.logger.info("suspending all VMs (--suspend)...")
 
             def _suspend(vm_name):
                 cls.logger.info(f"suspending VM '{vm_name}'...")
@@ -3839,7 +3919,8 @@ class BoxmanManager:
                 for vm_name, _ in vm_list
             ]
         else:
-            cls.logger.info("saving the state of all VMs to disk...")
+            if vm_list:
+                cls.logger.info("saving the state of all VMs to disk...")
 
             def _save(vm_name, workdir):
                 cls.logger.info(f"saving VM '{vm_name}' state to '{workdir}'...")
@@ -3853,6 +3934,9 @@ class BoxmanManager:
 
         [p.start() for p in processes]
         [p.join() for p in processes]
+
+        # Stop docker-compose clusters (keep containers; reversible via `up`).
+        cls.stop_compose_clusters()
 
         cls.logger.info("infrastructure is down")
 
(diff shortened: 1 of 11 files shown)


src/boxman/providers/docker_compose/session.py
```diff
@@ -0,0 +1,218 @@
...
+        return True
+
+    # -- internals ---------------------------------------------------------
+
+    def _compose_context(
+        self, cluster_name: str, cluster_cfg: dict[str, Any]
+    ) -> tuple[ComposeRunner, str, str]:
+        """Regenerate the compose file (idempotent) and build a runner.
+
+        Regenerating on every op keeps ``-f`` valid for down/stop/start
```
mherkazandjian: **Should-fix: teardown only works while the config still generates cleanly.**

`stop_cluster`/`down_cluster`/`destroy_cluster` all come through here, so every teardown re-runs `generate()` — which raises `ConfigError` for any box lacking `image:`/`build:` (compose_generator.py:135–139). That check is *not* performed by Phase-2 `normalize_v2_config`, so such a config loads fine and only explodes here. Concrete scenario: a provisioned cluster gets a new stub box (`cache: {}`) in conf.yml → `boxman down`/`deprovision` now exit 2 having torn down nothing, and `boxman destroy` (whose wrapper catches the error) nukes the workspace while the compose containers/networks stay orphaned.

Related side effects of regenerate-on-every-op:
- `destroy_cluster` writes the compose file (`write()` even `os.makedirs` the workdir) and then immediately deletes it.
- `boxman down` on a never-provisioned project *creates* the workdir + compose file.
- A hand-written `docker-compose.yml` in the cluster workdir is silently overwritten by any lifecycle op.
- `cluster_cfg["workdir"]` is a bare `KeyError` (not a `ConfigError`) when `workdir:` is missing — and it's the first line of every lifecycle op.

Suggestion: teardown ops should not need generation at all — `docker compose -p <project> stop|down` resolves containers from labels (compose v2), or fall back to the existing on-disk file when `generate()` raises. Keep regeneration for `up_cluster` only.

src/boxman/manager.py
```diff
@@ -3870,12 +3954,16 @@ def deprovision(cls, cli_args):
         # shared bridges before we touch libvirt state.
         cls.destroy_netlab()
 
+        # Tear down docker-compose clusters (`docker compose down`: remove
+        # containers + networks, keep named volumes) before libvirt state.
+        cls.deprovision_compose_clusters()
```
mherkazandjian: **Should-fix:** this call is unguarded and runs *before* the libvirt per-VM teardown, `destroy_networks()`, `deprovision_files()` and `unregister_from_cache()`. Since `down_cluster` → `_compose_context` → `generate()` can raise `ConfigError` (box without `image:`/`build:`), a broken dc cluster in a mixed project aborts the entire deprovision — the libvirt VMs stay up and the cache entry stays. Note `deprovision` is also invoked from `provision --force` (manager.py:3597), so the same failure blocks re-provisioning.

`destroy` already wraps its compose hook in try/except (step 2b, :4220–4224); this call deserves the same treatment (or move it after the libvirt teardown).

src/boxman/manager.py
```diff
@@ -3664,6 +3730,13 @@ def up(cls, cli_args):
         expected_vms = cls._get_project_vm_names()
 
         if not expected_vms:
+            # No libvirt VMs. A docker-compose-only project still has work to
+            # do: (idempotently) bring up its compose clusters. `up_cluster`
+            # is `docker compose up -d --wait`, which starts stopped containers
+            # and no-ops when everything is already running.
+            if cls._compose_clusters():
+                cls.provision_compose_clusters()
```
mherkazandjian: **Should-fix: first-run `up` on a dc-only project skips all provision bookkeeping.**

For a libvirt project, `up` with no existing VMs routes through full `provision()` (Case 1 below) — cache registration, `provision_files()` (cluster `files:` + runtime sentinels), netlab. This early-return branch handles both "first run" and "reconcile" for dc-only projects with a bare `provision_compose_clusters()`, so a dc-only project brought up via `boxman up` is never registered in the cache (the runtime-conflict tracking that `provision` relies on), its `files:` are never written, and a containerlab block is never deployed.

Suggestion: distinguish first-run (e.g. project not in cache) and route it through `cls.provision(cls, cli_args)` like Case 1 does — the compose part is idempotent either way.

src/boxman/providers/docker_compose/compose_generator.py
```diff
@@ -0,0 +1,198 @@
...
+import yaml
+
+from boxman import log
+from boxman.exceptions import ConfigError
+
+#: compose service keys copied through verbatim from a box definition
+_PASSTHROUGH_KEYS = (
+    "image",
+    "command",
+    "environment",
```
mherkazandjian, line 34: **Should-fix (promised-but-not-delivered): the Phase-2 templating deferral lands nowhere.**

`environment:`/`command:` are consumed here now, but `load_config`'s whole-file preprocessing (manager.py:419–445) still rewrites bare `{{ word }}` → `{word}` with no structure awareness, so `command: "echo {{ hostname }}"` in a dc box reaches the container as the literal `echo {hostname}`. PR #60's "Known deferral" and `doc/docker-compose-provider/config-schema.md:114` both say the structure-aware exemption is "deferred to Phase 3, where these values are actually consumed" — this is Phase 3, and neither an exemption nor a re-deferral (code comment / doc update) is in the diff. The Phase-2 caveat table also only covered `${VAR}`; nothing tests the `{{ word }}` case.

Minimum: update the doc to state which phase now owns it. Better: a cheap generator-side mitigation — warn when a passthrough `environment`/`command` value contains a `{word}` token (the corruption signature), pointing at the doc's rule of thumb.

src/boxman/providers/docker_compose/compose_generator.py
```diff
@@ -0,0 +1,198 @@
...
+        self,
+        cluster_name: str,
+        box_name: str,
+        box: dict[str, Any],
+        conf_dir: str,
+        cluster_networks: dict[str, Any],
+        shared: frozenset[str],
+    ) -> dict[str, Any]:
+        svc: dict[str, Any] = {}
+        for key in _PASSTHROUGH_KEYS:
```
mherkazandjian, line 139: **Should-fix: unknown box keys are silently dropped.**

Everything outside the 7 passthrough keys + `build`/`networks`/`volumes`/`compose_extra` vanishes with no diagnostic — and nothing upstream validates box keys either (`normalize_v2_config` only checks boxes/vms/provider shape). Two realistic casualties: `hostname:` (a *libvirt* box key users will carry over — and also a legitimate compose service key) and typos like `enviroment:`. That is inconsistent with this file's own warn+skip pattern for `volumes:` and shared-network refs, and `test_unknown_box_keys_are_dropped` currently enshrines the silence.

Suggestion: warn once per box for unconsumed keys ("unknown key 'X' — not part of the Phase-3 box schema; use compose_extra: if you need it in the service").

src/boxman/providers/docker_compose/session.py
```diff
@@ -0,0 +1,218 @@
...
+            f"wait ≤{timeout}s)"
+        )
+        runner.up(timeout)
+        return True
+
+    def stop_cluster(self, cluster_name: str, cluster_cfg: dict[str, Any]) -> bool:
+        """boxman down → ``docker compose stop`` (keep containers)."""
+        runner, _wd, _cf = self._compose_context(cluster_name, cluster_cfg)
+        self.logger.info(f"[{cluster_name}] docker compose stop")
+        runner.stop()
```
mherkazandjian: **Should-fix (small): teardown/stop results are never checked.**

`ComposeRunner.stop()/down()/down_volumes()` all run with `warn=True` and their `Result.ok` is ignored; `stop_cluster`/`down_cluster`/`destroy_cluster` return `True` unconditionally. With the docker daemon down, `boxman down` still ends with "infrastructure is down", and `destroy_cluster` deletes the generated compose file even when `down --volumes` just failed. `up_cluster` gets this right (the runner raises `ProvisionError`) — the teardown trio should at least `logger.warning` on `not result.ok`. Best-effort teardown is fine; *silent* best-effort is not.

src/boxman/manager.py
```diff
@@ -3642,6 +3704,10 @@ def provision(cls, cli_args):
         # display connection information (after ssh setup so connections are ready)
         cls.connect_info()
 
+        # bring up docker-compose clusters (no-op for libvirt-only projects);
+        # after libvirt VMs, mirroring the netlab hook's "extra infra last" order
+        cls.provision_compose_clusters()
```
mherkazandjian, line 3780: **Should-fix (small): failure here is a raw traceback.** `ComposeRunner.up` raises `ProvisionError` on failure/timeout and `preflight()` raises `RuntimeUnavailable`, but `main()` only converts `ConfigError` into a clean `log.error` + exit 2 (app.py:295–304). So the most ordinary operational failure on the new flagship path — a service that never turns healthy within `readiness_timeout` — hands the user a Python stack trace from `provision`/`up`. Either widen `main()`'s catch to `BoxmanError` (the message quality is already there) or wrap the compose hooks.

src/boxman/scripts/app.py
```diff
@@ -544,6 +544,18 @@ def _main():
...
+            # The docker-compose PROVIDER shells out to `docker compose` on
+            # the host; the `docker-compose` RUNTIME is libvirt-in-a-container
+            # (a different axis). Requiring runtime 'local' keeps them from
+            # being confused. Fail fast with a clean message.
+            if _ptype == 'docker-compose' and manager.runtime != 'local':
+                log.error(
+                    f"the docker-compose provider requires runtime 'local' "
+                    f"(got '{manager.runtime}'). The 'docker-compose' runtime "
+                    f"is libvirt-in-a-container — a different setting; see "
+                    f"doc/docker-compose-provider/config-schema.md."
```
mherkazandjian, line 560: Guardrail logic and placement are right (`manager.runtime` is set at :439, before this loop), but this pointer sends users to a doc that (a) never mentions the runtime-vs-provider axis — `config-schema.md` has zero occurrences of "runtime"; the axis is actually explained in `implementation-plan.md:24–25` — and (b) still contains "Not yet runnable: docker-compose / mixed configs … exits 2 … lands in Phase 3" (config-schema.md:78–85), which this PR falsifies. The same doc reference is baked into `DockerComposeSession._require_local_runtime`. Please update `config-schema.md` in this PR (drop/rewrite the "Not yet runnable" section, add the runtime-local requirement) or point at the doc that actually covers it.

One behavioural note worth documenting: because this fires for *every* subcommand, a project that pairs an existing libvirt-in-docker runtime with a newly added dc cluster can no longer run anything — including `destroy` — until the config is edited. Probably acceptable fail-fast, but it should be a documented consequence.

tests/test_docker_compose_provider.py
```diff
@@ -0,0 +1,544 @@
...
+        self.calls.append(("start", name))
+
+    def down_cluster(self, name, cfg):
+        self.calls.append(("down", name))
+
+    def destroy_cluster(self, name, cfg):
+        self.calls.append(("destroy", name))
+
+
+class TestManagerDispatch:
```
mherkazandjian, line 620: **Coverage gap:** these tests call the `*_compose_clusters()` helpers directly, so the 8 manager insertion points (provision :3709, up :3738/:3788/:3873, down :3939, deprovision :3959, destroy :4221, plus the down early-return guard :3902) are themselves untested — a future refactor dropping any one of them stays green, and the ordering guarantees the PR description advertises (netlab down → compose down → libvirt down; libvirt up → compose up → netlab up) aren't pinned anywhere. A couple of classmethod-level tests with a mocked session / `_get_vm_states` asserting call order would lock the seam in.

src/boxman/manager.py
```diff
@@ -2370,6 +2400,38 @@ def ensure_netlab_up(self) -> None:
...
+        """``docker compose up --wait`` every docker-compose cluster."""
+        for cluster_name, cluster in self._compose_clusters().items():
+            self.session_for_cluster(cluster_name).up_cluster(cluster_name, cluster)
+
+    def stop_compose_clusters(self) -> None:
+        """``docker compose stop`` every docker-compose cluster (boxman down)."""
+        for cluster_name, cluster in self._compose_clusters().items():
+            self.session_for_cluster(cluster_name).stop_cluster(cluster_name, cluster)
+
+    def start_compose_clusters(self) -> None:
```
mherkazandjian, line 2480: Nit: `start_compose_clusters` (and with it `DockerComposeSession.start_cluster` / `ComposeRunner.start`) has no non-test caller — the up-after-down path reconciles via `provision_compose_clusters()` (`up -d --wait`), which is arguably the better verb anyway. Either wire `start` into a flow that wants the cheaper no-recreate semantics, or note it as deliberate API surface for a later phase so it doesn't read as dead code.

src/boxman/manager.py
```diff
@@ -3854,6 +3935,9 @@ def _save(vm_name, workdir):
         [p.start() for p in processes]
         [p.join() for p in processes]
 
+        # Stop docker-compose clusters (keep containers; reversible via `up`).
+        cls.stop_compose_clusters()
```
mherkazandjian, line 4027: Nit: `down --suspend` reaches this same `docker compose stop` for dc clusters — there is no `docker compose pause` analog wired, so the flag is silently ignored for them. Acceptable for phase scope (control verbs are a later phase), but worth a line in the `down` docstring/help so `--suspend` users aren't surprised.

mherkazandjian (commented):
## Review — Phase 3 DockerComposeSession core lifecycle

Thorough pass over the full diff and the actual code at `911b5b0` (plus the base branch for Phase 1/2 context). **No blockers**, but a cluster of teardown-robustness issues that should be fixed in this PR or immediately after. The two headline claims were stress-tested:

**Claim 1 — "libvirt flows stay byte-identical": verified, holds.** All ~20 `config['clusters']` → `_vm_clusters` conversions are per-VM / per-network flows; for a libvirt-only project `_vm_clusters` yields the same clusters mapping (same nested objects, same order), so behaviour is unchanged — including in-place mutations (`_resolve_iso_config`) which still hit `self.config` through the shared nested dicts. The four remaining `self.config['clusters']` sites (`provision_files` manager.py:1124, `deprovision_files` :1152, the `write_ssh_config` VM count :3024, `_select_vm_targets` :4428) are dc-safe (`.get('vms', {})` or files/workdir-only). `normalize_v2_config` keeps dc `boxes:` verbatim, `_vm_cluster_map`/`_get_project_vm_names` already use `.get('vms')`, and dc-only `provision` cannot hang in the IP wait (`get_connect_info` returns True for zero VMs). The `down` message-gating changes are no-ops for libvirt-only projects.

**Verified non-issue (recording it since it was the first thing I checked):** `DockerComposeSession._conf_dir()` relying on `manager.config_path` is correct in production — `BoxmanManager.__init__` sets `self.config_path = config` (manager.py:42, :78–79), app.py constructs the manager with the conf path (app.py:429) and assigns `session.manager = manager` for every session (app.py:591). The test fixture (`SimpleNamespace(runtime=..., config_path=...)`) mirrors reality; `build.context` resolution against the conf.yml dir (D4) works.

**Claim 2 — dc lifecycle correctness:** the up/provision side is correct and idempotent — all four `up` code paths plus the provision hook are wired, and ordering vs netlab matches the description. The weak side is **teardown**: every teardown op re-generates (re-validates + re-writes) the compose file, so a config that no longer generates cleanly cannot be torn down, and `deprovision`'s unguarded compose hook can block libvirt teardown in a mixed project. Details inline.

### Should-fix
1. **Teardown requires a re-generable config** — `session.py:125–146`: `stop_cluster`/`down_cluster`/`destroy_cluster` all run `generate()` (raises `ConfigError` for a box without `image:`/`build:` — a check Phase-2 `normalize_v2_config` does NOT perform) and `write()` before touching docker. Broken box ⇒ `down`/`deprovision` exit 2 having torn down nothing; `destroy` logs-and-continues but orphans the compose containers/networks while nuking the workspace. Also: `destroy_cluster` writes the file then deletes it, and `down` on a never-provisioned project creates the workdir + compose file as a side effect.
2. **`deprovision_compose_clusters()` unguarded, placed before libvirt teardown** — manager.py:3959 (contrast destroy's wrapped step 2b at :4220–4224). A `ConfigError` there aborts VM, network, files and cache teardown for the whole project; `deprovision` is also called from `provision --force` (:3597).
3. **dc-only `boxman up` first-run bypasses provision bookkeeping** — manager.py:3732–3739: no `register_project_in_cache()`, no `provision_files()` (cluster `files:` never written), no netlab. A libvirt project's first `up` routes through full `provision`; the dc-only branch should too.
4. **Phase-2-promised templating exemption not delivered nor re-deferred** — bare `{{ word }}` inside compose `environment:`/`command:` values is still corrupted to `{word}` by `load_config` (manager.py:419–445); PR #60's "Known deferral" and `doc/docker-compose-provider/config-schema.md:114` defer it to "Phase 3, where these values are actually consumed" — this is Phase 3, it consumes them, and neither code nor doc was updated.
5. **Unknown box keys silently dropped** — compose_generator.py:111: anything outside the 7 passthrough keys + `build`/`networks`/`volumes`/`compose_extra` vanishes without a trace (`hostname:`, `entrypoint:`, or a typo like `enviroment:`), inconsistent with the volumes/shared-networks warn+skip pattern; no schema layer catches it either.
6. **Teardown results never checked** — session.py:86–121 / compose_runner.py:59–73: `stop()`/`down()`/`down_volumes()` run with `warn=True`, `.ok` is never inspected, the session methods return True unconditionally; `boxman down` prints "infrastructure is down" with the docker daemon down, and destroy deletes the compose file even when `down --volumes` failed.
7. **`ProvisionError`/`RuntimeUnavailable` surface as tracebacks on the flagship path** — `main()` only converts `ConfigError` to a clean exit 2 (app.py:295–304); an unhealthy service (`ComposeRunner.up` raise) or missing docker (preflight) hands the user a Python stack trace from `provision`/`up`.
8. **The docs the new errors point at are stale/wrong** — the guardrail (app.py:556) and the session error direct users to `config-schema.md` for the runtime-vs-provider distinction, but that doc never mentions the runtime axis, and its "Not yet runnable: docker-compose … exits 2 … lands in Phase 3" section (:78–85) is falsified by this very PR. The diff touches no docs.
9. **Manager wiring untested** — tests/test_docker_compose_provider.py:479: `TestManagerDispatch` calls the `*_compose_clusters()` helpers directly, but nothing drives `provision`/`up`/`down`/`deprovision`/`destroy` to pin the 8 insertion points or their ordering; deleting any hook keeps the suite green.

### Nits
10. `start_compose_clusters`/`start_cluster`/`ComposeRunner.start` have no non-test caller (manager.py:2420) — up-after-down reconciles via `provision_compose_clusters()`. Wire it or mark it as future API surface.
11. `down --suspend` silently degrades to `docker compose stop` for dc clusters (manager.py:3906–3939) — fine for phase scope, but undocumented (`docker compose pause` would be the analog).
12. `_vm_clusters` is a `@property` recomputed on every access (calling `provider_type_for_cluster` per cluster) while `_compose_clusters()` is a method — cheap in practice, but inconsistent and read ~20×/operation.
13. `_require_local_runtime`'s `getattr(self.manager, "runtime", "local")` default means the "session driven directly" case it exists for silently passes when `manager` is unset (session.py:164); similarly `_compose_context`'s `cluster_cfg["workdir"]` raises a bare `KeyError` instead of a friendly `ConfigError` when `workdir:` is missing (session.py:134).

Solid phase overall: clean generator/runner/session separation, good unit coverage of the new units themselves, honest phase-scoping warnings, and the coarse-dispatch seam genuinely leaves libvirt untouched.

src/boxman/manager.py
```diff
@@ -233,6 +233,36 @@ def session_for_vm(self, full_vm_name: str) -> Optional["ProviderSession"]:
...
+            for name, cluster in (self.config.get('clusters') or {}).items()
+            if not self._is_compose_cluster(name)
+        }
+
+    def _compose_clusters(self) -> dict[str, Any]:
+        """docker-compose clusters, in config order (empty for libvirt-only projects)."""
+        return {
+            name: cluster
+            for name, cluster in (self.config.get('clusters') or {}).items()
+            if self._is_compose_cluster(name)
```
mherkazandjian, line 265: **[P1] Gate Docker Compose dispatch on schema v2.** `_apply_config_version()` explicitly warns versionless/v1 users that `boxes:` are ignored, but this collection still includes those clusters and the new generator turns the boxes into live services. I reproduced a versionless config logging that warning while provisioning its declared service. Reject or exclude Docker Compose clusters under v1 so the compatibility promise and actual behavior agree.

src/boxman/providers/docker_compose/compose_generator.py
```diff
@@ -0,0 +1,198 @@
...
+            raise ConfigError(
+                f"box '{cluster_name}.{box_name}' must define 'image:' or "
+                f"'build:'."
+            )
+        return svc
+
+    def _build(self, build: Any, conf_dir: str) -> Any:
+        """Resolve ``build.context`` to an absolute path vs *conf_dir* (D4)."""
+        if isinstance(build, str):
+            return os.path.abspath(os.path.join(conf_dir, os.path.expanduser(build)))
```
mherkazandjian: **[P2] Preserve remote build contexts.** Compose accepts Git/URL contexts in both scalar `build:` and mapping `build.context`, but this unconditional path resolution turns `https://github.com/docker/awesome-compose.git#main` into `/project/https:/github.com/docker/awesome-compose.git#main`. Resolve only relative filesystem contexts, or reject remote contexts explicitly if they are intentionally unsupported. The mapping branch below has the same fault.

src/boxman/manager.py
```diff
@@ -4124,6 +4212,17 @@ def destroy(cls, cli_args):
             except Exception as exc:
                 cls.logger.warning(f"deprovision raised: {exc} — continuing")
 
+        # 2b. fully tear down docker-compose clusters — destroy goes beyond
+        #     deprovision's `docker compose down` (keeps named volumes) to
+        #     `down --volumes` and removes the generated compose file. Runs
```
mherkazandjian, line 4333: **[P1] Disclose named-volume deletion before confirmation.** This path executes `down --volumes`, which permanently removes named Docker volumes that users can already declare through `compose_extra`, but the pre-prompt plan lists only VMs, networks, generated files, runtime state, workspace, and templates. Please add an explicit per-cluster warning about named-volume deletion before asking for confirmation.

src/boxman/manager.py
```diff
@@ -2472,7 +2534,7 @@ def vm_clone_tasks():
         # directory; doing it here sequentially prevents the race condition
         # where N parallel processes all try to define the same pool at once.
         seen_workdirs: set = set()
-        for cluster_name, cluster in self.config['clusters'].items():
+        for cluster_name, cluster in self._vm_clusters.items():
```
mherkazandjian, line 2608: **[P1] Use the owning libvirt session when creating this storage pool.** In a valid Compose-primary mixed project, `self.provider` is `DockerComposeSession`; `_ensure_libvirt_storage_pool()` then reads the Compose provider config, loses the libvirt URI/runtime settings, and may make `VirshCommand` fall back to local `qemu:///system`. I reproduced the helper receiving the Compose config instead of the configured remote libvirt URI. Pass `cluster_name` or its session into the helper and use `session_for_cluster(cluster_name).provider_config`.

src/boxman/providers/docker_compose/session.py
```diff
@@ -0,0 +1,218 @@
...
+    def _conf_dir(self) -> str:
+        """Directory of the project ``conf.yml`` (for build-context resolution)."""
+        config_path = getattr(self.manager, "config_path", None)
+        return os.path.dirname(os.path.abspath(config_path)) if config_path else os.getcwd()
+
+    def _compose_project(self, cluster_name: str) -> str:
+        """Derive the ``docker compose -p`` name — one project per cluster
+        (ADR-001), so clusters never share compose state."""
+        base = self._provider_config.get("project_name") or self.config.get("project") or "boxman"
+        return _sanitize_project_name(f"{base}_{cluster_name}")
```
mherkazandjian, line 263: **[P1] Make the Compose project identity collision-resistant.** Distinct cluster names such as `web.api` and `web_api` both sanitize to `lab_web_api` here; case-only differences collide as well. Those clusters then share the same Compose labels/state, and teardown with `--remove-orphans` can delete the sibling cluster’s containers. Append a stable digest of the unsanitized project/cluster identity or reject collisions before provisioning.

src/boxman/manager.py
```diff
@@ -4124,6 +4212,17 @@ def destroy(cls, cli_args):
             except Exception as exc:
                 cls.logger.warning(f"deprovision raised: {exc} — continuing")
 
+        # 2b. fully tear down docker-compose clusters — destroy goes beyond
+        #     deprovision's `docker compose down` (keeps named volumes) to
+        #     `down --volumes` and removes the generated compose file. Runs
+        #     regardless of the libvirt-in-container runtime state: the
+        #     compose provider shells out to the host docker directly.
+        try:
+            cls.destroy_compose_clusters()
```
mherkazandjian, line 4337: **[P1] This hook can be skipped by the preceding “nothing to do” return.** That predicate does not consider Compose clusters, their workdirs/artifacts, or Docker project state. I reproduced an existing dc cluster workdir with an empty cache and no `workspace.path`: `destroy` reported “nothing to do” and never called this method. Include Compose state or cluster workdirs in the predicate so teardown remains retryable after cache loss.

Orski174:
## Review addressed — `9925ed1`

Thanks for the thorough pass. All 9 should-fix + both P1/P2 rounds + the 4 nits are addressed in `9925ed1` (1240 host unit tests green, +31). Disposition below; the correctness and teardown items also got a live nested run inside stage01 (never the host).

### Correctness / mixed projects
| # | Finding | Resolution |
|---|---|---|
| P1 | v1 configs spin up dc services | `_reject_v1_docker_compose` in `_apply_config_version` — v1/versionless + any dc cluster → `ConfigError` at load. |
| P1 | storage pool uses dc default session | `_ensure_libvirt_storage_pool(workdir, cluster_name)` now builds `VirshCommand` from `session_for_cluster(cluster_name).provider_config`. Pinned by a unit test (dc primary, distinct libvirt URI). |
| P1 | project-name collisions | `_reject_compose_project_collisions()` runs before provisioning; `web.api`/`web_api` → `ConfigError`. |
| should-fix 3 | dc-only first-run `up` skips bookkeeping | first run (not in cache) routes through full `provision()`; reconcile stays idempotent. **Live-verified.** |

### Teardown robustness
| # | Finding | Resolution |
|---|---|---|
| should-fix 1 | teardown re-generates config | new `_teardown_runner`: uses on-disk file if present, else label-only (`docker compose -p <project> …`); `up_cluster` is the only regenerator. Broken/absent config tears down; hand-written files respected; never-provisioned `down` creates nothing. **Live-verified** (label-based `down` of a non-generating config). |
| should-fix 2 | `deprovision` hook unguarded/before libvirt | wrapped best-effort (like destroy's 2b). |
| should-fix 6 / C2 | teardown results unchecked | `_check` warns on `not ok`; destroy keeps the file when `down --volumes` fails. |
| P1 | destroy "nothing to do" skips compose | predicate now includes on-disk compose state. **Live-verified** (destroy retryable after cache loss). |

### Error surfacing & UX
| # | Finding | Resolution |
|---|---|---|
| should-fix 7 | tracebacks on flagship path | `main()` catches `BoxmanError` → exit 2. |
| C1 / C3 | `workdir` KeyError / `readiness_timeout` ValueError | `_workdir` and `_readiness_timeout` raise `ConfigError`. |
| P1 | named-volume deletion undisclosed | destroy plan lists a per-cluster `down --volumes` line. **Live-verified.** |

### Generator fidelity
| # | Finding | Resolution |
|---|---|---|
| P2 | remote build contexts mangled | `_is_remote_build_context` — Git/URL/`github.com/…` preserved verbatim. |
| should-fix 5 | unknown box keys dropped silently | warn once per box for keys outside the schema. |
| should-fix 4 | `{{ word }}` templating deferral | generator warns on the `{word}` corruption signature + doc updated. |
| C4 | `use_sudo` ignored | `ComposeRunner` honors it. |

### Docs & tests
| # | Finding | Resolution |
|---|---|---|
| should-fix 8 | `config-schema.md` stale | dropped the falsified "Not yet runnable" section; added the provider-vs-runtime axis, `runtime: local` requirement, v1 rejection, templating status. |
| should-fix 9 | manager wiring untested | `TestFlowWiring` drives provision/up(×3)/down/deprovision/destroy with a stubbed session and asserts the hooks fire in order. |

### Nits
10 `start_*` marked reserved control-verb API · 11 `--suspend` dc behavior documented in `down` · 12 `_compose_clusters` made a `@property` (symmetry with `_vm_clusters`) · 13 `_require_local_runtime` manager-unset default explained.

One item scoped down: the compose-primary mixed **storage-pool** fix (P1) is pinned by a deterministic unit test rather than a full nested-KVM provision — a real mixed libvirt+dc e2e is disproportionate here and fits Phase 4's hybrid e2e better. Happy to run it if you'd prefer.

PR #62, opened 2026-07-16
feat(docker-compose): Phase 4 networking — macvlan attach to shared bridges (VM↔container L2)

Orski174:
Phase 4 of the docker-compose provider epic (#42). Closes #52.

## What this does
Lets a docker-compose container and a libvirt VM share an L2 domain by
attaching the container to the same host Linux bridge as the VM, via a
docker **macvlan** network whose `parent` is that bridge.

## Changes
- **Generator** — `shared_networks` refs become top-level macvlan networks
  (`driver: macvlan` + `driver_opts.parent` + `ipam` from
  `subnet`/`gateway`/`ip_range`). Box `networks:` supports the list form
  (auto address) and the mapping form (`{net: {ipv4_address: …}}`, static).
  Missing `bridge`/`subnet` on an attached shared net → `ConfigError`;
  `ipv4_address` on a cluster-internal net → warn + drop.
- **Netfilter (D8)** — default `disable_netfilter: false`: keep the
  host-global `bridge-nf-call-iptables` untouched and install idempotent
  per-bridge scoped physdev accept rules on `FORWARD` (+ `DOCKER-USER` when
  present). Global disable is an explicit, loudly-warned opt-in.
- **Manager** — dc-only reconcile `up` ensures shared bridges before the
  macvlan containers reconcile (survives a host reboot dropping the bridge).
- **Docs** — `config-schema.md` shared_networks + static-IP syntax;
  `design.md` D8 marked implemented, macvlan example corrected.

## Testing
- **Host unit suite:** 1251 passed (net +8 over Phase 3), integration deselected.
- **Live hybrid e2e in stage01** (nested, per the host-vs-VM policy — never the
  host hypervisor). One conf.yml: a libvirt VM (`node01`, eth1 → `br-p4app`) +
  a docker-compose container (`web`, macvlan on `br-p4app` static 10.10.0.10 +
  cluster-internal `backend`). `boxman provision` → `destroy`:
  - **AC10** VM↔container ping both ways — 3/3 each direction
  - **AC11** ARP both ways — MACs cross-verified (real L2, not routing)
  - **AC12** cluster-internal `backend` isolated from the VM — 100% loss
  - D8 rules present on `FORWARD` + `DOCKER-USER`; `bridge-nf-call-iptables=1`
    left untouched
  - `destroy` tears down both clusters; shared bridge left in place by design

## Notes
- The dc-only **reconcile** bridge-ensure path is pinned by unit test, not
  live-exercised (consistent with Phase 3's precedent).
- Live dogfooding re-confirmed two known prereq issues (guest-agent check
  bypasses the `-c` URI; `sshpass` missing breaks key injection) — already on
  the epic's candidate-issues list, not addressed here.

### DIFF
--- a/boxes/docker-compose-standalone/README.md
+++ b/boxes/docker-compose-standalone/README.md
@@ -35,9 +35,9 @@ boxman provision
 # nginx answers on the published port
 curl -sI localhost:8080 | head -1              # HTTP/1.1 200 OK
 
-# inspect the generated, hand-runnable compose file
-docker compose -f .boxman/web/docker-compose.yml ps
-cat .boxman/web/docker-compose.yml
+# inspect the stack — boxman runs it under project 'dc_standalone_web'
+docker compose -p dc_standalone_web ps
+cat .boxman/web/docker-compose.yml     # emits a top-level 'name:', so `-f <file> ps` also works
 
 # redis is healthy behind the frontend
 docker exec dc_standalone_web-cache-1 redis-cli ping   # PONG
--- a/boxes/docker-compose-standalone/README.md
+++ b/boxes/docker-compose-standalone/README.md
@@ -1,0 +1,63 @@
+# docker-compose-standalone
+
+The simplest **docker-compose provider** example — no libvirt, no VMs. Two
+containers on one cluster-internal bridge network:
+
+```
+frontend (nginx, :8080)  --depends_on: service_healthy-->  cache (redis)
+```
+
+On `boxman provision` boxman translates the `boxes:` into a
+`docker-compose.yml` (written to the cluster `workdir`, `./.boxman/web/`) and
+runs `docker compose up -d --wait`. It shows the core docker-compose surface:
+`image`, `ports`, a `healthcheck`, a `depends_on` with a
+`condition: service_healthy` (passed through verbatim), and a cluster-internal
+bridge network with an explicit `subnet:`.
+
+## Prerequisites
+
+- Docker Engine + the **compose v2** plugin (`docker compose version`).
+- boxman on the **`local` runtime** (the default). The docker-compose provider
+  shells out to `docker compose` on the host; it is a config error under the
+  `docker` runtime.
+- Free TCP port **8080** on the host.
+
+## Bring it up
+
+```bash
+cd boxes/docker-compose-standalone
+boxman provision
+```
+
+## Verify
+
+```bash
+# nginx answers on the published port
+curl -sI localhost:8080 | head -1              # HTTP/1.1 200 OK
+
+# inspect the generated, hand-runnable compose file
+docker compose -f .boxman/web/docker-compose.yml ps
+cat .boxman/web/docker-compose.yml
+
+# redis is healthy behind the frontend
+docker exec dc_standalone_web-cache-1 redis-cli ping   # PONG
+```
+
+`boxman up` is idempotent — re-running reconciles (starts anything stopped and
+re-asserts health).
+
+## Tear down
+
+```bash
+boxman down       # docker compose stop (keeps containers/volumes)
+boxman up         # bring them back
+boxman destroy    # docker compose down --volumes + remove the generated file
+```
+
+## Notes
+
+- The generated `docker-compose.yml` lives under `.boxman/` (git-ignored). It
+  is a fidelity artifact — you can run `docker compose -f ... <cmd>` against it
+  directly.
+- Need a compose feature boxman does not model yet? Add it under
+  `compose_extra:` (per-box or per-cluster) and it is deep-merged verbatim.
--- a/boxes/hybrid-libvirt-docker-compose/README.md
+++ b/boxes/hybrid-libvirt-docker-compose/README.md
@@ -41,9 +41,8 @@ host-unreachable `10.10.0.x` address for node01's management IP.
   with `ssh-copy-id` over a password session (`apt install sshpass` /
   `dnf install sshpass`). Without it, key injection is skipped (non-fatal) and
   you fall back to the guest agent for the checks below.
-- The macvlan support this box relies on lands with **Phase 4** — run it from a
-  checkout that has it (the `feat/docker-compose-provider` line, once #62 is
-  merged).
+- The macvlan support this box relies on requires the
+  `feat/docker-compose-provider` line (the docker-compose provider epic).
 - First run downloads the Ubuntu 24.04 cloud image (~0.5 GB, cached under
   `~/.cache/boxman/images`).
 
@@ -109,7 +108,11 @@ shared bridges can be used by multiple projects, so removing them is an explicit
 action:
 
 ```bash
-sudo ip link del bx_app          # only if nothing else uses it
+# remove the two scoped ACCEPT rules ensure() installed for this bridge...
+sudo iptables -D FORWARD -i bx_app -o bx_app -m physdev --physdev-is-bridged -j ACCEPT
+sudo iptables -D DOCKER-USER -i bx_app -o bx_app -m physdev --physdev-is-bridged -j ACCEPT  # if the chain exists
+# ...then the bridge itself (only if nothing else uses it)
+sudo ip link del bx_app
 ```
 
 ## Troubleshooting
--- a/boxes/hybrid-libvirt-docker-compose/README.md
+++ b/boxes/hybrid-libvirt-docker-compose/README.md
@@ -19,10 +19,15 @@ On `boxman up` boxman:
 1. creates the host bridge `bx_app` and installs a **scoped** per-bridge
    netfilter accept rule (decision D8 — `bridge-nf-call-iptables` is left
    untouched host-wide),
-2. clones/boots `node01` with its second NIC on `bx_app` (static `10.10.0.20`
-   via a per-MAC netplan match — the shared bridge has no DHCP),
+2. clones/boots `node01` with its second NIC (`adapter_2`) attached to `bx_app`,
 3. generates the docker-compose file and brings `web` up: a macvlan network on
    `bx_app` (static `10.10.0.10`) plus a cluster-internal `backend` bridge.
+
+`node01`'s management IP is the DHCP address on its mgmt NIC (that is what
+`boxman ssh` uses). The shared bridge is a plain L2 domain with **no DHCP**, so
+`node01`'s address on it (`10.10.0.20`) is assigned as an explicit step in the
+test below — deliberately not at boot, so boxman doesn't mistake that
+host-unreachable `10.10.0.x` address for node01's management IP.
 
 ## Prerequisites
 
@@ -61,32 +66,35 @@ The container is `hybrid_libvirt_dc_services-web-1` (find it with
 `docker ps --filter name=web`). It has `10.10.0.10` on the shared bridge and a
 `backend` address on `172.31.0.0/24`.
 
-**Ping both ways** (VM `10.10.0.20` ⇄ container `10.10.0.10`):
+**Step 1 — give node01 its address on the shared bridge.** It has no DHCP there,
+so assign it once (the app NIC is the second one, `enp7s0` on this box):
 
 ```bash
-# container -> VM
-docker exec hybrid_libvirt_dc_services-web-1 ping -c3 10.10.0.20
-
-# VM -> container
-boxman ssh compute_node01            # or: ssh -F <workspace>/ssh_config compute_node01
-node01$ ping -c3 10.10.0.10
+boxman ssh compute_node01 -- sudo ip addr add 10.10.0.20/24 dev enp7s0
+boxman ssh compute_node01 -- ip -4 -br addr show enp7s0     # -> 10.10.0.20/24
 ```
 
-**ARP resolves across the bridge** (each side learns the other's real MAC —
-proof this is L2, not routed):
+**Step 2 — ping both ways** (VM `10.10.0.20` ⇄ container `10.10.0.10`):
 
 ```bash
-node01$ ip neigh show 10.10.0.10                 # -> lladdr <container mac> REACHABLE
+docker exec hybrid_libvirt_dc_services-web-1 ping -c3 10.10.0.20   # container -> VM
+boxman ssh compute_node01 -- ping -c3 10.10.0.10                   # VM -> container
+```
+
+**Step 3 — ARP resolves across the bridge** (each side learns the other's real
+MAC — proof this is L2, not routed):
+
+```bash
+boxman ssh compute_node01 -- ip neigh show 10.10.0.10       # -> lladdr <container mac>
 docker exec hybrid_libvirt_dc_services-web-1 ip neigh show 10.10.0.20
 ```
 
-**Cluster-internal isolation** — the container's `backend` address is *not*
-reachable from the VM (only the shared bridge is L2-adjacent):
+**Step 4 — cluster-internal isolation.** The container's `backend` address is
+*not* reachable from the VM (only the shared bridge is L2-adjacent):
 
 ```bash
-# find the container's backend IP
-docker exec hybrid_libvirt_dc_services-web-1 ip -4 addr show eth0
-node01$ ping -c2 -W2 172.31.0.<that ip>          # -> 100% loss (isolated)
+docker exec hybrid_libvirt_dc_services-web-1 ip -4 -br addr   # note the 172.31.0.x
+boxman ssh compute_node01 -- ping -c2 -W2 172.31.0.2          # -> 100% loss (isolated)
 ```
 
 ## Tear down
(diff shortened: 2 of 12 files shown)


tests/test_netlab_shared_bridges.py
```diff
@@ -83,34 +83,82 @@ def fake_run(cmd, **kwargs):
...
+        cfg = {"lab_mgmt": {"bridge": "br1", "disable_netfilter": True}}
+        with patch("boxman.netlab.shared_bridges.run",
+                   side_effect=self._fake_run()) as run:
+            with patch("pathlib.Path.exists", return_value=True):
+                with caplog.at_level(_logging.WARNING, logger="boxman"):
+                    shared_bridges.ensure(cfg)
+        all_cmds = " | ".join(c.args[0] for c in run.call_args_list)
+        assert "bridge-nf-call-iptables" in all_cmds and "echo 0" in all_cmds
+        assert "-I FORWARD" not in all_cmds  # global disable, no scoped rule
+        assert any("HOST-WIDE" in r.message for r in caplog.records)
```
mherkazandjian: **[blocking] This test fails deterministically — standalone and in the full suite.**

`boxman`'s logger is non-propagating (`src/boxman/loggers/logger.py:76` sets `logger.propagate = False`), and pytest's capture handler hangs off the *root* logger — so `caplog.at_level(_logging.WARNING, logger="boxman")` raises the level but `caplog.records` stays empty, and the final `assert any("HOST-WIDE" …)` is always `False` (the warning is visibly emitted to stdout via boxman's own handler, yet never captured).

Reproduced on a clean checkout of this head (pytest 9.0.2):
- `pytest tests/test_netlab_shared_bridges.py` → **1 failed, 14 passed**
- `pytest tests/ -m "not integration"` → **1 failed, 1250 passed** — the PR body's "1251 passed" doesn't reproduce.

Also worth noting: this repo has no CI check that runs pytest (only the Copilot reviewer), so nothing will catch this at merge time.

The project already ships the exact remedy: the `captured_logs` fixture in `tests/conftest.py`, whose docstring names this precise `propagate = False` gotcha (it flips propagation for the test and restores it). Ten+ existing tests use it with the same `rec.message` assertion pattern. I applied the change below locally and the file passes 15/15:

```suggestion
    def test_disable_netfilter_opt_in_sets_global_and_warns(self, captured_logs):
        """Explicit disable_netfilter: true → host-global sysctl=0, a loud
        warning, and NO scoped FORWARD rule."""
        cfg = {"lab_mgmt": {"bridge": "br1", "disable_netfilter": True}}
        with patch("boxman.netlab.shared_bridges.run",
                   side_effect=self._fake_run()) as run:
            with patch("pathlib.Path.exists", return_value=True):
                shared_bridges.ensure(cfg)
        all_cmds = " | ".join(c.args[0] for c in run.call_args_list)
        assert "bridge-nf-call-iptables" in all_cmds and "echo 0" in all_cmds
        assert "-I FORWARD" not in all_cmds  # global disable, no scoped rule
        assert any("HOST-WIDE" in r.message for r in captured_logs.records)
```

mherkazandjian (commented):
## Review — Phase 4 macvlan networking (verified at head `36efcb1`)

Reviewed with the branch checked out; findings were verified against the real code and, where possible, empirically (unit suite executed, `docker compose` project-name derivation probed live, spike `poc.sh` topology cross-checked). **Verdict: the design and implementation are solid — the generator, the D8 scoped rules, and the bridges-before-compose reconcile ordering all check out — but one shipped unit test fails deterministically and should be fixed before merge.** Everything else is minor.

### Ranked findings

**Blocking**
1. `tests/test_netlab_shared_bridges.py:148` — `test_disable_netfilter_opt_in_sets_global_and_warns` fails deterministically, standalone **and** in the full suite: raw `caplog` cannot see the `boxman` logger (`propagate = False`, `src/boxman/loggers/logger.py:76`). On a clean checkout of this head: `pytest tests/ -m "not integration"` → **1 failed, 1250 passed** (pytest 9.0.2) — the PR body's "1251 passed" does not reproduce. One-word fix (the project's own `captured_logs` fixture), verified locally → file passes 15/15. There is no pytest CI check on this repo, so nothing else will catch it. Details inline.

**Should fix**
2. `compose_generator.py:318` — a scalar `networks: app_bridge` (forgotten `[…]`) is silently normalized to *no networks*; compose then attaches the service to the project-**default** bridge, so `up -d --wait` succeeds while the macvlan L2 attach silently never happened. A scalar per-network value (`networks: {app_bridge: 10.10.0.5}`) raises a raw `ValueError` instead of `ConfigError`. (Copilot flagged the silent-drop half; I verified both, plus that no upstream schema guard exists — `normalize_v2_config` leaves dc `boxes:` verbatim.)
3. `boxes/docker-compose-standalone/README.md:39` — the verify step `docker compose -f .boxman/web/docker-compose.yml ps` targets the **wrong compose project** and lists nothing (empirically confirmed: without `-p`, compose derives `name: web` from the file's directory; the runner labels everything `dc_standalone_web`). Fix the command, and consider emitting top-level `name:` from the generator to make the D5 "hand-runnable" claim true.
4. `boxes/hybrid-libvirt-docker-compose/conf.yml:17` — header still says the VM's shared-bridge IP is "assigned by a per-MAC netplan match in cloud-init"; the cloud-init below deliberately does not do that (its own comment + README make it a manual step). Stale from a previous iteration.

**Nits**
5. `shared_bridges.py:50` — the scoped rule is right to install, but the docstring attributes it to the VM↔macvlan-container pairing, which per the spike's own topology data is the one path that likely never traverses filter `FORWARD` (details inline; scenario 1's proven "blocked" baseline was veth-port↔veth-port, and the macvlan scenarios had no negative control).
6. `hybrid README:112` — teardown documents removing the bridge but not the two scoped ACCEPT rules `ensure()` installed for it.
7. *(no inline)* Two boxes pinning the same `ipv4_address` on one shared net — or an address outside the declared `subnet` — surface as a runtime `docker compose up` error rather than a generate-time `ConfigError`. Optional cheap check in `_service_networks`.
8. *(no inline, pre-existing)* `doc/docker-compose-provider/README.md` still says "Phase 0: Design & Validation (this phase)"; and hybrid README line 44's "lands with **Phase 4** … once #62 is merged" will read stale the moment this merges — consider "requires the `feat/docker-compose-provider` line".

### Verified clean (checked and explicitly refuted)
- **Reconcile ordering (`manager.py:3821`)**: every `up` path that reaches `provision_compose_clusters` runs `ensure_shared_bridges` first — dc-only reconcile (new), all-VMs-running (pre-existing), start/resume (pre-existing at 3895); first-run/partial route through `provision()` which calls it at 3702. The three ordering tests pin this correctly, and `ensure()` is idempotent (`-C`-before-`-I`, `ip link show` guard).
- **`_shared_macvlan_networks` KeyError safety**: every name in `referenced_shared` provably passed `_require_macvlan_ipam` against the same dict object (a `None` entry raises there as `{}` → missing `bridge`), so `entry["subnet"]`/`entry["bridge"]` cannot KeyError.
- **README container names**: `dc_standalone_web-cache-1` / `hybrid_libvirt_dc_services-web-1` are exactly right for `_compose_project()` (`<project_name>_<cluster>`) + compose-v2 `<project>-<service>-<n>` naming.
- **FORWARD + DOCKER-USER dual insertion**: redundancy is by-design and holds up — FORWARD works on docker-less hosts; the DOCKER-USER copy survives daemon restarts (spike 4b); post-restart, docker's re-inserted jump sits above the demoted FORWARD copy but jumps straight into DOCKER-USER where the ACCEPT is at position 1.
- **Rule/bridge teardown asymmetry**: a dangling `-i bx_app` rule after a manual bridge delete is an inert no-op and re-arms correctly if the name is reused — consistent with the documented bridges-outlive-projects policy (just document it, nit 6).
- **`networks: {backend: null}`** → service `{backend: {}}` mapping form is valid compose; the mixed-form output is test-pinned.
- **Determinism**: macvlan networks emit in `shared_networks` declaration order, services in `boxes:` order.
- **`compose_extra` overriding `driver`/`parent` on the macvlan net**: documented D7 escape-hatch semantics ("deep-merged verbatim, extra wins"), test-pinned — acceptable.
- **Unquoted `{{ env("BOXMAN_ADMIN_PASS") }}`** in the hybrid conf: established convention in 10+ existing boxes — not this PR's issue.
- Remaining "Phase 4" mentions in `design.md` / `implementation-plan.md` / `spike/findings.md` are accurate historical attributions, not stale scope statements.

Good PR overall — the D8 scoped-rules work in particular is careful and well-evidenced. Fix the test and this is mergeable.

src/boxman/providers/docker_compose/compose_generator.py
```diff
@@ -231,27 +248,91 @@ def _service_networks(
...
+        List form yields empty opts per ref; mapping form carries each ref's
+        option dict (``{ipv4_address: …}``, or ``None`` → empty).
+        """
+        if not networks:
+            return []
+        if isinstance(networks, dict):
+            return [(ref, dict(opts or {})) for ref, opts in networks.items()]
+        if isinstance(networks, (list, tuple)):
+            return [(ref, {}) for ref in networks]
+        return []
```
mherkazandjian: **[should-fix] Silent drop of a scalar `networks:` — and a raw `ValueError` for scalar per-network options.**

Two under-validated inputs in `_normalize_box_networks`, and this method is the only chokepoint (`normalize_v2_config` keeps dc `boxes:` verbatim — there is no upstream schema guard):

1. **Bare string** — `networks: app_bridge` (a user forgetting the list brackets) hits this `return []` and is treated as *no networks*, with no warning. The consequence is nastier than a missing attach: the service is emitted without a `networks:` key at all, so compose attaches it to the project-**default** bridge network — `docker compose up -d --wait` succeeds, the container reports healthy, and the macvlan L2 attach silently never happened. You only find out when pings fail. That's much quieter than the adjacent failure modes, which warn loudly (unknown ref) or raise `ConfigError` (missing `bridge`/`subnet`).
2. **Scalar opts** — in the dict branch above, `dict(opts or {})` assumes each value is a mapping or `None`. `networks: {app_bridge: 10.10.0.5}` (a very plausible guess at the static-IP syntax) crashes with a raw `ValueError: dictionary update sequence element #0 …` instead of a `ConfigError`.

Suggested shape:

```python
if isinstance(networks, str):
    return [(networks, {})]          # or: raise ConfigError(...) — anything but silence
if isinstance(networks, dict):
    for ref, opts in networks.items():
        if opts is not None and not isinstance(opts, dict):
            raise ConfigError(
                f"box '{cluster_name}.{box_name}': networks['{ref}'] must be a "
                f"mapping like {{ipv4_address: …}} or null (got {opts!r}).")
    return [(ref, dict(opts or {})) for ref, opts in networks.items()]
if isinstance(networks, (list, tuple)):
    return [(str(ref), {}) for ref in networks]
raise ConfigError(f"box '{cluster_name}.{box_name}': 'networks:' must be a list or mapping (got {type(networks).__name__}).")
```

(The `cluster_name`/`box_name` params are already threaded in but currently unused — this would put them to work.)

boxes/docker-compose-standalone/README.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+## Verify
+
+```bash
+# nginx answers on the published port
+curl -sI localhost:8080 | head -1              # HTTP/1.1 200 OK
+
+# inspect the generated, hand-runnable compose file
+docker compose -f .boxman/web/docker-compose.yml ps
```
mherkazandjian: **[should-fix] This verify command targets the wrong compose project and prints nothing.**

Without `-p`, compose v2 derives the project name from the compose *file's directory* — empirically confirmed against this exact layout: `docker compose -f …/web/docker-compose.yml config` resolves to `name: web`. But boxman created the stack under `-p dc_standalone_web` (`session._compose_project()` → `<project_name>_<cluster>`, and `ComposeRunner` always passes `-p`). Since `ps` filters containers by the project label, this command shows an **empty list** while the stack is happily up — the opposite of what a verify step should do.

Two fixes, ideally both:
- **README (minimal):** `docker compose -p dc_standalone_web ps` (the `-f` becomes optional). The same wrong command is in the conf.yml header comment (line 22).
- **Generator (durable):** emit top-level `name: <compose project>` in `ComposeGenerator.generate()`. Compose v2 honors it, `-p` still overrides it (runner behavior unchanged), and it would make the module's D5 claim — "inspectable and hand-runnable with `docker compose -f <workdir>/docker-compose.yml ps`" — actually true for every generated file instead of only when the right `-p` is remembered.

(The other verify steps are correct — I checked `dc_standalone_web-cache-1` against the project-name derivation and compose-v2 `<project>-<service>-<n>` naming.)

boxes/hybrid-libvirt-docker-compose/conf.yml
```diff
@@ -0,0 +1,156 @@
...
+#
+#   node01 (libvirt)                          web (docker-compose)
+#     adapter_1 -- dhcp --- mgmt NAT            eth0 -- backend (172.31.0.0/24)
+#     adapter_2 -- 10.10.0.20/24 --\            eth1 -- 10.10.0.10 (macvlan) --\
+#                                   \                                          \
+#                                    `------ bx_app (shared bridge, 10.10.0.0/24) -'
+#
+# boxman creates bx_app (+ a scoped netfilter accept rule — decision D8),
+# attaches node01's second NIC to it, and brings the macvlan container up on
+# it. The VM's shared-bridge address is assigned by a per-MAC netplan match
```
mherkazandjian: **[should-fix, docs] Stale header: there is no "per-MAC netplan match in cloud-init" in this box.**

This sentence contradicts the actual design shipped in this file: the cloud-init block below contains no netplan/network config — its own comment says the shared-bridge address is "deliberately NOT here" — and the README makes the assignment an explicit manual step (`boxman ssh compute_node01 -- sudo ip addr add 10.10.0.20/24 dev enp7s0`). Looks like a leftover from the earlier iteration this head commit replaced ("explicit shared-bridge IP").

Suggested rewording: "The VM's shared-bridge address is assigned as an explicit step in the README (the shared bridge is a plain L2 domain with no DHCP)."

src/boxman/netlab/shared_bridges.py
```diff
@@ -42,6 +42,50 @@ def _set_sysfs(path: str, value: str) -> None:
     run(f"echo {value} | sudo tee {path}", hide=True)
 
 
+def _scoped_rule_body(bridge: str) -> str:
+    """The physdev accept rule body shared by ``-C`` (check) and ``-I`` (insert).
+
+    Accepts frames both entering and leaving *bridge* that are bridged
+    (``--physdev-is-bridged``) — i.e. L2 lab traffic between a bridge-attached
+    VM and a macvlan-attached container on the same bridge — so it isn't
```
mherkazandjian: **[nit, docstring precision] The "i.e." names the one pairing this rule likely never sees.**

The rule itself is right to install and the mechanics are sound — but `-i <br> -o <br> -m physdev --physdev-is-bridged` matches frames the bridge **forwards between two ports** (br_netfilter's FORWARD traversal). A docker-macvlan endpoint is not a bridge port: it's a child of the bridge *device*, so on mainline kernels VM→container frames leave the bridge via the local pass-up path (delivered to the macvlan rx handler on the bridge device), and container→VM frames enter via `br_dev_xmit` (bridge LOCAL_OUT) — neither traverses iptables filter `FORWARD`/`DOCKER-USER` at all.

The spike's own data is consistent with this: the proven "blocked" baseline (scenario 1) used two **veth ports** (`spike-vm`/`spike-ct` netns stand-ins), while the macvlan scenarios (6a–6c) were only ever run with the scoped rule *already installed* — there's no macvlan-without-rule negative control. So the demonstrated load-bearing case for this rule is **port↔port** traffic on the shared bridge (VM↔VM, VM↔containerlab veth — which docker's FORWARD DROP genuinely breaks), and AC10/AC11 would very likely have passed even without it.

No behavior change needed — two cheap follow-ups:
1. Reword the "i.e." to something like "e.g. VM↔VM / bridge-port peers on the shared bridge (and belt-and-braces for macvlan-attached containers)".
2. Next spike run: add a `nf=1, no rule` macvlan negative control to pin the kernel behavior on record.

(Corollary worth knowing: since the VM↔macvlan path bypasses iptables, it also can't be *filtered* via `DOCKER-USER` — relevant if anyone later asks for per-bridge firewalling of that traffic.)

boxes/hybrid-libvirt-docker-compose/README.md
```diff
@@ -0,0 +1,127 @@
...
+boxman destroy
+```
+
+This removes `node01`, the mgmt network, and the docker-compose cluster
+(`down --volumes`). The shared bridge `bx_app` is **left in place by design** —
+shared bridges can be used by multiple projects, so removing them is an explicit
+action:
+
+```bash
+sudo ip link del bx_app          # only if nothing else uses it
```
mherkazandjian: **[nit, docs] Teardown covers the bridge but not the scoped rules that came with it.**

`ensure()` installed two persistent ACCEPT rules for this bridge (`FORWARD` and `DOCKER-USER`), and nothing ever removes them. After `ip link del bx_app` they linger referencing a nonexistent interface — inert and harmless, and they'd deliberately re-arm if the bridge name is reused — but they *are* host-global state boxman added, and a user following this cleanup note will still have them. Suggest one line alongside the bridge removal:

```bash
sudo iptables -D FORWARD -i bx_app -o bx_app -m physdev --physdev-is-bridged -j ACCEPT
sudo iptables -D DOCKER-USER -i bx_app -o bx_app -m physdev --physdev-is-bridged -j ACCEPT   # if the chain exists
```

(Also worth a sentence in config-schema.md's netfilter note, which currently documents rule installation but not their lifetime.)

PR #64, opened 2026-07-18
feat(docker-compose): Phase 5 volumes — named/bind/workdir mounts (#53)

Orski174:
Phase 5 of the docker-compose provider epic (#42). Closes #53.

## What this does
Translates a box's structured `volumes:` (until now warn+skipped) into compose
mounts. The lifecycle (down keeps volumes / destroy removes them) was already
wired in Phase 3 via the runner's `down` vs `down --volumes`, so this PR is
generation + bind-dir pre-creation + the `size:` decision.

## Volume schema
A list of structured mounts on a box; `container_path` is required, and a
`host_path` decides the kind:

| Kind | Trigger | Emitted | Top-level `volumes:` |
|---|---|---|---|
| **named** | has `name`, no `host_path` | `<name>:<container_path>[:ro]` | `{<name>: {driver: local}}` |
| **bind** | has `host_path` (relative → vs conf.yml dir, D4) | `<abs host>:<container_path>[:ro]` | — |
| **workdir** | a bind with `host_path: .` | `<conf dir>:<container_path>[:ro]` | — |

- `readonly: true` → `:ro`; `compose_extra:` on a named entry deep-merges onto
  its top-level spec.
- **`size:` is advisory** (agreed design decision) — docker's `local` driver
  can't enforce quotas, so boxman emits the volume + warns rather than
  pretending to cap it (`compose_extra: {driver_opts: …}` stays available for
  setups that can).
- Malformed input fails fast with `ConfigError` (non-list `volumes:`,
  non-mapping entry, missing `container_path`, named entry missing `name`)
  rather than silently dropping a mount.

## Bind-dir handling
`session._ensure_bind_dirs` `mkdir -p`s each bind host dir before `up` (as the
boxman user, so docker doesn't create it `root`-owned). Named volumes are
docker-managed; bind dirs are **never** removed on teardown (they're user
paths).

## Testing
- **Host unit:** 1269 passed (`pytest -m "not integration"`), +13 volume cases
  (emission for all three kinds, readonly, advisory/ignored size, shared named
  volume defined once, `compose_extra` merge, the four `ConfigError` paths,
  unknown-key warn, bind-dir pre-creation).
- **Live e2e in stage01** (dc-only, pure docker): a service with a named volume
  + a bind mount →
  - **AC13** named volume survives `down`/`up` (marker file intact)
  - **AC14** bind mount works both directions (host↔container)
  - **AC15** `destroy` removes the named volume and **leaves** the bind dir + its files
  - bind dir pre-created `admin:admin` (not root)

## Notes
- A committed volumes **example box** is deferred to Phase 8 (#56), consistent
  with the plan; this PR used a throwaway conf for the live e2e.

### DIFF
--- a/src/boxman/providers/docker_compose/compose_generator.py
+++ b/src/boxman/providers/docker_compose/compose_generator.py
@@ -5,11 +5,20 @@ ComposeGenerator — translate a boxman docker-compose *cluster* into a
 Scope: services (image / build / command / environment / ports /
 depends_on / restart / healthcheck), cluster-internal bridge networks,
 ``shared_networks`` bridges attached as docker **macvlan** networks (Phase 4
-— L2 to libvirt VMs on the same host bridge), and the ``compose_extra:``
-escape hatch (deep-merged verbatim, per-cluster and per-box — design
-decision D7). Out-of-phase box features are warned about and skipped:
+— L2 to libvirt VMs on the same host bridge), structured ``volumes:`` (Phase
+5 — named / bind / workdir mounts), and the ``compose_extra:`` escape hatch
+(deep-merged verbatim, per-cluster and per-box — design decision D7).
 
-- ``volumes:`` (structured named/bind/workdir mounts) → **Phase 5**.
+A box's ``volumes:`` is a list of structured mounts:
+
+- **named** (``{name: pg_data, container_path: /var/lib/postgresql/data}``) —
+  emitted as ``pg_data:/var/lib/postgresql/data`` plus a top-level
+  ``volumes: {pg_data: {driver: local}}``. An optional ``size:`` is
+  **advisory** (warned) — docker's local driver does not enforce quotas.
+- **bind** (``{host_path: ./configs, container_path: /etc/app,
+  readonly: true}``) — emitted as ``<abs host>:/etc/app:ro``; a relative
+  ``host_path`` is resolved against the project ``conf.yml`` dir (D4). A
+  "workdir mount" is just a bind mount with ``host_path: .``.
 
 A box's ``networks:`` may be a plain list (``[app_bridge, backend]`` — the
 container gets an auto-assigned address on each) or a mapping that pins a
@@ -49,15 +58,24 @@ _PASSTHROUGH_KEYS = (
 #: before the brace), which is a safe compose interpolation, not corruption.
 _CORRUPTED_TEMPLATE_RE = re.compile(r"(?<!\$)\{[a-zA-Z_]\w*\}")
 
-#: every box key the Phase-3 generator understands. Anything else is warned
-#: about and dropped (``volumes:`` → Phase 5 is handled separately, but is a
-#: recognised key so it doesn't also trip the unknown-key warning).
+#: every box key the generator understands. Anything else is warned about and
+#: dropped.
 _KNOWN_BOX_KEYS = frozenset(_PASSTHROUGH_KEYS) | {
     "build",
     "networks",
     "volumes",
     "compose_extra",
 }
+
+#: keys understood inside a structured ``volumes:`` entry. Others warn.
+_KNOWN_VOLUME_KEYS = frozenset({
+    "name",
+    "container_path",
+    "host_path",
+    "readonly",
+    "size",
+    "compose_extra",
+})
 
 
 class ComposeGenerator:
@@ -106,12 +124,16 @@ class ComposeGenerator:
         #: only these become top-level macvlan networks (insertion order via
         #: shared_networks below keeps the output deterministic).
         referenced_shared: set[str] = set()
+        #: named volumes defined by any box → the top-level ``volumes:`` block,
+        #: in first-seen order (a name shared by two boxes is defined once).
+        named_volumes: dict[str, Any] = {}
 
         services: dict[str, Any] = {}
         for box_name, box in (cluster_cfg.get("boxes") or {}).items():
             services[box_name] = self._service(
                 cluster_name, box_name, box or {}, conf_dir,
                 cluster_networks, shared_networks, referenced_shared,
+                named_volumes,
             )
 
         compose: dict[str, Any] = {}
@@ -124,6 +146,8 @@ class ComposeGenerator:
         )
         if networks:
             compose["networks"] = networks
+        if named_volumes:
+            compose["volumes"] = named_volumes
 
         # D7: per-cluster escape hatch, deep-merged verbatim.
         return self._deep_merge(compose, cluster_cfg.get("compose_extra") or {})
@@ -148,6 +172,7 @@ class ComposeGenerator:
         cluster_networks: dict[str, Any],
         shared_networks: dict[str, Any],
         referenced_shared: set[str],
+        named_volumes: dict[str, Any],
     ) -> dict[str, Any]:
         svc: dict[str, Any] = {}
 
@@ -176,12 +201,11 @@ class ComposeGenerator:
         if nets:
             svc["networks"] = nets
 
-        if box.get("volumes"):
-            self.logger.warning(
-                f"box '{cluster_name}.{box_name}': 'volumes:' is not supported "
-                f"yet (lands in Phase 5) — skipping {len(box['volumes'])} "
-                f"volume(s). Use 'compose_extra:' if you need them now."
-            )
+        vols = self._service_volumes(
+            cluster_name, box_name, box.get("volumes"), conf_dir, named_volumes
+        )
+        if vols:
+            svc["volumes"] = vols
 
         # D7: per-box escape hatch, deep-merged verbatim (last, so it can
         # override anything boxman generated).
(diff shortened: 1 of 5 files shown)


src/boxman/providers/docker_compose/compose_generator.py
```diff
@@ -254,6 +278,85 @@ def _build(self, build: Any, conf_dir: str) -> Any:
...
+                )
+            container_path = entry.get("container_path")
+            if not container_path:
+                raise ConfigError(
+                    f"box '{cluster_name}.{box_name}': a 'volumes:' entry is "
+                    f"missing 'container_path' ({entry!r})."
+                )
+            ro = ":ro" if entry.get("readonly") else ""
+            host_path = entry.get("host_path")
+            if host_path:
```
mherkazandjian, line 356: nit: on a bind entry, `compose_extra:` is silently ignored — it's in `_KNOWN_VOLUME_KEYS` (so the unknown-key warning never fires; that warning even *recommends* `compose_extra:`), but it's only consumed in the named-volume branch below. A co-present `name:` is likewise silently dropped here (`host_path` decides the kind per the docs, but silently). Given `size:`-on-bind gets an explicit "ignored" warning just above, these two deserve the same one-line courtesy — or a doc note that `compose_extra:`/`name:` are named-volume-only (config-schema.md currently only says what `compose_extra:` does "on a named entry", not that it's a no-op on a bind).

src/boxman/providers/docker_compose/compose_generator.py
```diff
@@ -254,6 +278,85 @@ def _build(self, build: Any, conf_dir: str) -> Any:
...
+                if entry.get("size"):
+                    self.logger.warning(
+                        f"box '{cluster_name}.{box_name}': 'size:' is ignored on "
+                        f"the bind mount for '{container_path}' — it is only "
+                        f"meaningful on a named volume."
+                    )
+                abs_host = self._abs_context(str(host_path), conf_dir)
+                mounts.append(f"{abs_host}:{container_path}{ro}")
```
mherkazandjian, line 373: optional hardening: with short-syntax emission, a `:` in `container_path` (or in `host_path`, or in a volume `name` — line 353 too) silently changes the meaning of the emitted string, and a non-absolute `container_path` only fails later at `docker compose up` with a docker-level error pointing at the generated file. Given this module's stated fail-fast stance ("malformed input raises ConfigError rather than silently dropping a mount"), a cheap config-time check (no `:` in path/name, `container_path` absolute) would surface those as a `ConfigError` naming the box. Fine to punt — the inputs are pathological — just noting it since the fail-fast contract is otherwise thorough.

src/boxman/providers/docker_compose/session.py
```diff
@@ -248,6 +249,44 @@ def _conf_dir(self) -> str:
...
+        """
+        conf_dir = self._conf_dir()
+        for box_name, box in (cluster_cfg.get("boxes") or {}).items():
+            for entry in (box or {}).get("volumes") or []:
+                if not isinstance(entry, dict):
+                    continue  # malformed — the generator raises on it
+                host_path = entry.get("host_path")
+                if not host_path:
+                    continue  # named volume — docker-managed
+                abs_host = os.path.abspath(
```
mherkazandjian: nit (DRY): this expression is a character-for-character re-implementation of `ComposeGenerator._abs_context`, which the generator uses for the *same* `host_path` in `_service_volumes`. The docs promise bind resolution follows "the same rule as `build.context`", and that guarantee currently rests on two copies in two files staying identical — if one ever changes, `_ensure_bind_dirs` would mkdir a different path than the one docker mounts. The session already holds `self._generator`, so reusing the helper (or hoisting it to a shared module-level function) makes divergence impossible.

mherkazandjian (commented):
Reviewed the full generation path (`_service_volumes`), the session's bind-dir pre-creation (`_ensure_bind_dirs`), the docs, and the new tests, against full-file context on the head branch.

**Verified working as intended**
- `volumes` is in `_KNOWN_BOX_KEYS`, so a box with `volumes:` no longer trips the unknown-key warning.
- Ordering in `up_cluster` is right: `_compose_context` runs `generate()` — which raises `ConfigError` on malformed volumes — and writes the file *before* `_ensure_bind_dirs`, so bind dirs are only created for a config that already generated cleanly; the `continue  # malformed` guard is purely defensive, and creation is race-safe (`exist_ok=True`) with a graceful `OSError` fallback.
- Path resolution: absolute and `~`-prefixed `host_path`s resolve correctly through `_abs_context` (`os.path.join` discards `conf_dir` when the expanded path is absolute); relative paths resolve against the conf.yml dir exactly like `build.context`, as documented.
- `name:/path:ro` is valid compose short syntax for a readonly named volume; the top-level `volumes:` block is deterministic (first-seen order, insertion-ordered dicts, `sort_keys=False`).
- Lifecycle claims match the Phase-3 wiring: `down` keeps named volumes, `destroy` runs `down --volumes` and removes the generated file only on success.
- The 13 new tests are sound and cover all three mount kinds, readonly, advisory/ignored `size:`, shared-name-defined-once, `compose_extra` merge, all four `ConfigError` paths, the unknown-key warning, and bind-dir pre-creation (the `session.os` patches only affect `_ensure_bind_dirs`, since `_compose_context` is stubbed).

**Verdict:** solid, well-scoped, well-tested — nits only, nothing blocking. Three small inline notes below.

**+1 to Copilot's two comments** (both verified still present on the head):
- `config-schema.md` line 86 still says "Still out of scope: structured `volumes:` (Phase 5), …", now contradicted by the new `## Volumes` section — worth fixing in this PR.
- Shared named volume: first-seen-wins means a second box's differing `compose_extra:` for the same `name` is silently discarded; a one-line divergence warning would fit this module's loud-warning style. (`size:` is unaffected — its advisory warning fires per-entry, and size is never recorded anyway.)

**Optional, non-blocking:** a non-mapping `compose_extra:` on a volume entry surfaces as a raw `AttributeError` from `_deep_merge` rather than a `ConfigError` — but that matches the pre-existing box/cluster/network-level `compose_extra` handling, so fine to punt to a shared fix.

src/boxman/providers/docker_compose/compose_generator.py
```diff
@@ -254,6 +278,85 @@ def _build(self, build: Any, conf_dir: str) -> Any:
...
+                )
+            unknown = [k for k in entry if k not in _KNOWN_VOLUME_KEYS]
+            if unknown:
+                self.logger.warning(
+                    f"box '{cluster_name}.{box_name}': ignoring unknown "
+                    f"'volumes:' key(s) {', '.join(repr(k) for k in unknown)} — "
+                    f"use 'compose_extra:' to pass extra mount options."
+                )
+            container_path = entry.get("container_path")
+            if not container_path:
```
mherkazandjian, line 335: 💡 Minor: worth also validating that `container_path` is absolute here. A relative `container_path` passes generation and only fails at `docker compose` deploy time with a much less clear error. One more `ConfigError` branch next to this one would fail fast at generation.

src/boxman/providers/docker_compose/compose_generator.py
```diff
@@ -254,6 +278,85 @@ def _build(self, build: Any, conf_dir: str) -> Any:
...
+                        f"mount) — {entry!r}."
+                    )
+                if entry.get("size"):
+                    self.logger.warning(
+                        f"box '{cluster_name}.{box_name}': size: "
+                        f"{entry['size']!r} on named volume '{name}' is advisory "
+                        f"— docker's local driver does not enforce quotas."
+                    )
+                mounts.append(f"{name}:{container_path}{ro}")
+                if name not in named_volumes:
```
mherkazandjian, line 397: 💡 Minor: when two boxes share a named volume and both set `compose_extra`, the first-seen spec silently wins (order-dependent). A warning when a later box's `compose_extra` differs from the recorded one would make the sharing explicit.

src/boxman/providers/docker_compose/session.py
```diff
@@ -248,6 +249,44 @@ def _conf_dir(self) -> str:
...
+                    os.path.join(conf_dir, os.path.expanduser(str(host_path)))
+                )
+                if os.path.exists(abs_host):
+                    continue
+                try:
+                    os.makedirs(abs_host, exist_ok=True)
```
mherkazandjian, line 282: 💡 Minor edge case: a single-file bind mount (e.g. `./app.conf:/etc/app.conf`) whose host path doesn't exist yet gets pre-created as a *directory* here (docker would do the same, as root). Intent can't be distinguished from the schema, but worth a note in `config-schema.md` that missing bind paths are always created as directories.

src/boxman/providers/docker_compose/session.py
mherkazandjian: 💡 Nit: this duplicates the generator's `_abs_context` path resolution (`abspath(join(conf_dir, expanduser(p)))`). Extracting one shared helper would keep the two implementations from drifting apart.

mherkazandjian (approved):
Reviewed in depth: read the full diff, ran the full test suite on this branch (1269 passed, 126 deselected — matches the PR description), and verified the `up_cluster` ordering (`_compose_context` generates + validates the compose file before `_ensure_bind_dirs` runs, so malformed `volumes:` entries raise `ConfigError` before any `mkdir` happens). Lifecycle semantics (down keeps / `down --volumes` removes) are consistent with the Phase 3 wiring, and the docs match the code.

Approving — four minor suggestions inline, none blocking.

PR #65, opened 2026-07-18
fix: make hybrid-libvirt-containerlab boxes deploy (netlab path + cisco_vios + shared-bridge)

Orski174:
## What & why

The three `hybrid-libvirt-containerlab-*` boxes never deployed end-to-end. Boundary-testing the Cisco box in a nested-KVM staging VM surfaced **four** distinct blockers before the (license-gated) vendor image even matters. This PR fixes all four and adds a regression test.

## Fixes

| # | Scope | Fix |
|---|-------|-----|
| **A** — netlab path bug | boxman code — **all 3 boxes** | `deploy_netlab`/`ensure_netlab_up` resolved the startup-config against `os.path.dirname(self.config_path)`; the documented `cd boxes/<box> && boxman up` passes the relative default `conf.yml`, so that's `""` → `render_topology` falls back to the workspace dir → `FileNotFoundError`. Now uses `os.path.dirname(os.path.abspath(...))`. + regression tests. |
| **B** — Cisco kind/image | cisco box | `cisco_vios_l2` isn't a containerlab kind → `kind: cisco_vios` + `type: switch`, `image: vrnetlab/cisco_vios:L2-20200929`. |
| **C** — Cisco interface | cisco box | `sw1:Gi0/1` (IOS slot notation) invalid for `cisco_vios` → flat `sw1:eth1`. |
| **D** — shared-bridge attach | **all 3 boxes** | `host:clab_mgmt` *creates* a host interface and collides with the shared bridge. Declare `clab_mgmt` as a `kind: bridge` node and link `sw1:<port> ↔ clab_mgmt:eth-sw1`. |

A and D are feature-wide; B and C are Cisco-only (Arista/Nokia already used valid kinds + flat interface names).

## Verification

- `86` netlab unit tests pass; the 2 new regression tests fail on the pre-fix code (`source_root == ""`) and pass after.
- In a nested-KVM staging VM, `boxman netlab deploy` (documented relative-`conf.yml` flow) now renders the topology and clears **all** containerlab topology validation (kind, type, bridge node, link), stopping **only** at the image pull:
  `Failed to pull image docker.io/vrnetlab/cisco_vios:L2-20200929 ... repository does not exist` — i.e. the licensed image the user builds/supplies.
- The libvirt half (template reuse → host01 boot → DHCP → SSH) was independently confirmed working.

> Note: full end-to-end (switch actually boots, L2 forwarding works, `eth1`↔`Gi0/1` mapping) still requires a licensed Cisco IOSvL2 qcow2 and cannot be verified in CI.

## Testing on stage01 (nested-KVM staging VM)

Prereqs inside the VM: docker, libvirt/qemu-kvm (host-passthrough for nested KVM), and containerlab (`bash -c "$(curl -sL https://get.containerlab.dev)"`).

**1. Boundary test — no vendor image (proves A+B+C+D):**
```bash
cd boxes/hybrid-libvirt-containerlab-cisco-iosvl2
export BOXMAN_CISCO_ADMIN_PASS=admin BOXMAN_CISCO_ENABLE_PASS=admin
sudo ip link add clab_mgmt type bridge && sudo ip link set clab_mgmt up   # boxman up does this itself
boxman netlab deploy
# EXPECT: topology renders + validates; fails only at
#   "Failed to pull image .../cisco_vios:L2-20200929 ... repository does not exist"
```

**2. Full end-to-end — with a licensed image:**
```bash
# build the vendor image once (needs your Cisco IOSvL2 qcow2)
git clone https://github.com/srl-labs/vrnetlab.git
cp vios_l2-...-20200929.qcow2 vrnetlab/cisco/vios/     # vios_l2 filename => switch build
( cd vrnetlab/cisco/vios && make )                      # => vrnetlab/cisco_vios:L2-20200929
docker images | grep cisco_vios

cd boxes/hybrid-libvirt-containerlab-cisco-iosvl2
boxman up                       # builds Ubuntu template + host01, bridge, deploys sw1
boxman netlab inspect
$(boxman netlab ssh sw1)        # drop into IOS: sw1#
boxman ssh host01               # host01's 2nd NIC should get a DHCP lease from sw1 (192.0.2.0/24)
boxman destroy
sudo ip link delete clab_mgmt   # shared bridges are not auto-removed
```

**Free alternative (no Cisco license):** the sibling `hybrid-libvirt-containerlab-arista-ceos` box exercises the same A+D fixes with a downloadable cEOS image.

### DIFF
--- a/src/boxman/manager.py
+++ b/src/boxman/manager.py
@@ -2392,7 +2392,15 @@ class BoxmanManager:
         if netlab is None:
             return
         netlab.preflight()
-        source_root = os.path.dirname(self.config_path) if self.config_path else None
+        # Resolve startup-config templates relative to the *config file's*
+        # directory. abspath() is required: run the documented way
+        # (`cd boxes/<box> && boxman up`), config_path is the bare relative
+        # default "conf.yml", so os.path.dirname() would return "" — which
+        # render_topology treats as "unset" and falls back to the workspace
+        # dir, where configs/ does not exist (FileNotFoundError on sw1's
+        # startup-config).
+        source_root = (os.path.dirname(os.path.abspath(self.config_path))
+                       if self.config_path else None)
         netlab.render_topology(source_root=source_root)
         netlab.deploy()
 
@@ -2427,8 +2435,11 @@ class BoxmanManager:
             return
         netlab.preflight()
         # Render the topology so ensure_up has a .clab.yml to deploy from
-        # if the lab is missing entirely.
-        source_root = os.path.dirname(self.config_path) if self.config_path else None
+        # if the lab is missing entirely. abspath() so a bare relative
+        # --conf (the default "conf.yml") resolves to the box directory
+        # rather than "" — see deploy_netlab() for the full rationale.
+        source_root = (os.path.dirname(os.path.abspath(self.config_path))
+                       if self.config_path else None)
         netlab.render_topology(source_root=source_root)
         netlab.ensure_up()
 
--- a/boxes/hybrid-libvirt-containerlab-arista-ceos/conf.yml
+++ b/boxes/hybrid-libvirt-containerlab-arista-ceos/conf.yml
@@ -159,6 +159,14 @@ containerlab:
         kind: arista_ceos
         image: ceos:4.32.0F
         startup-config: configs/sw1.cfg.j2
+      # A `kind: bridge` node references the *existing* shared Linux
+      # bridge that boxman creates from `shared_networks.clab_mgmt` (and
+      # that host01's second NIC also joins) — it attaches, it does not
+      # create one. (`host:<name>` would create a new host interface and
+      # collide with the bridge.)
+      clab_mgmt:
+        kind: bridge
     links:
-      # sw1's Ethernet1 lands on the same host bridge as host01's second NIC.
-      - endpoints: ["sw1:eth1", "host:clab_mgmt"]
+      # sw1's Ethernet1 lands on the same host bridge as host01's second
+      # NIC; eth-sw1 is the bridge-side veth on the shared bridge.
+      - endpoints: ["sw1:eth1", "clab_mgmt:eth-sw1"]
--- a/boxes/hybrid-libvirt-containerlab-cisco-iosvl2/README.md
+++ b/boxes/hybrid-libvirt-containerlab-cisco-iosvl2/README.md
@@ -40,32 +40,35 @@ Cisco Modeling Labs / Enterprise Agreement are the usual paths.
 
   ```bash
   git clone https://github.com/srl-labs/vrnetlab.git
-  cd vrnetlab/cisco/vios_l2
+  cd vrnetlab/cisco/vios
   ```
 
-**3. Copy the qcow2 into the vios_l2 directory and build.**
+**3. Copy the qcow2 into the `cisco/vios` directory and build.**
 
-  vrnetlab's Makefile derives the image tag from the qcow2 filename,
-  so rename the file to encode the version cleanly first.
+  srl-labs/vrnetlab builds both the router and the L2 switch from this
+  one directory and auto-selects the switch personality from a
+  `vios_l2` filename. The Makefile derives the image tag from the
+  filename, so keep the `vios_l2-` prefix and a clean date/version.
 
   ```bash
-  # rename to make the version unambiguous
+  # keep the vios_l2 prefix so the build detects the L2 switch;
+  # the trailing date becomes the image tag (L2-<date>)
   cp /path/to/vios_l2-adventerprisek9-m.SSA.high_iron_20200929.qcow2 \
-     vios_l2-15.2.2020T.qcow2
+     vios_l2-20200929.qcow2
 
   # build the vrnetlab container (takes a few minutes; boots the qcow2
   # once to snapshot the post-init state)
   make
 
   # verify
-  docker images | grep cisco_vios_l2
-  # vrnetlab/cisco_vios_l2   15.2.2020T   <id>   <time>   ~600MB
+  docker images | grep cisco_vios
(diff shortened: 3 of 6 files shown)


src/boxman/manager.py
```diff
@@ -2392,7 +2392,15 @@ def deploy_netlab(self) -> None:
...
         netlab.preflight()
-        source_root = os.path.dirname(self.config_path) if self.config_path else None
+        # Resolve startup-config templates relative to the *config file's*
+        # directory. abspath() is required: run the documented way
+        # (`cd boxes/<box> && boxman up`), config_path is the bare relative
+        # default "conf.yml", so os.path.dirname() would return "" — which
+        # render_topology treats as "unset" and falls back to the workspace
+        # dir, where configs/ does not exist (FileNotFoundError on sw1's
+        # startup-config).
+        source_root = (os.path.dirname(os.path.abspath(self.config_path))
```
mherkazandjian, line 2402: Verified this is the right fix and that it's complete:

- `render_topology()` resolves its default with a truthiness check (`Path(source_root) if source_root else self.workdir`), so the pre-fix `""` from `dirname("conf.yml")` silently fell back to the workspace dir — `abspath()` first is exactly the right cure, and absolute or nested-relative `--conf` values behave as before (`None` still short-circuits to `None`).
- The only other `dirname(self.config_path ...)` site (the `netlab` property's workdir fallback) is already guarded by `or "."`, so this and its twin in `ensure_netlab_up()` cover all affected sites.

Optional follow-up (non-blocking): the root cause is that `render_topology()` conflates `""` with "unset". A defensive normalization there — treat only `None` as unset, or raise/absolutize on an empty string — would keep a future call site from reintroducing this silently.

mherkazandjian (commented):
Reviewed with full-file context on the head branch (`src/boxman/manager.py`, `src/boxman/netlab/containerlab.py`, all three box configs). All four fixes hold up and the regression tests are real — LGTM, non-blocking notes only.

**A — netlab source root: correct and complete.** `render_topology()` resolves its default with a truthiness check (`Path(source_root) if source_root else self.workdir`), so the pre-fix `os.path.dirname("conf.yml") == ""` silently redirected startup-config lookup to the workspace dir, where `configs/` doesn't exist — exactly the reported `FileNotFoundError`. `dirname(abspath(...))` pins it to the box dir under the documented `cd boxes/<box>` flow; absolute and nested-relative `--conf` values resolve as before, and `None` still short-circuits to `None`. I checked for other sites with the same pattern: the only other `dirname(self.config_path ...)` (the `netlab` property's workdir fallback) is already guarded by `or "."`, so the two fixed sites cover everything affected.

**D — shared-bridge attach: verified the name chain in each box.** `shared_networks.clab_mgmt.bridge: clab_mgmt` (the literal host bridge device) ← host01's `adapter_2` with `network_source: clab_mgmt` ← containerlab `clab_mgmt: {kind: bridge}` node + `sw1:<port> ↔ clab_mgmt:eth-sw1` link. The bridge-kind node attaches to the existing bridge instead of creating a colliding `host:` interface, so sw1 and host01 land on one L2 segment. Applied consistently across the arista/cisco/nokia boxes; `eth-sw1` is a valid host-side veth name (7 chars, well under IFNAMSIZ) and unique within each topology.

**Tests: non-vacuous.** `mgr._netlab = fake` only short-circuits the cached getter, so `deploy_netlab`/`ensure_netlab_up` still execute the real `source_root` computation and the fake's `render_topology` captures it; the assertions (truthy, absolute, realpath == cwd) fail on the pre-fix `""`. Nice touch using `realpath` on both sides.

**One observation on staging coverage (non-blocking):** the PR states the staging run "clears all containerlab topology validation (kind, type, bridge node, link), stopping only at the image pull". To be precise: containerlab pulls images before creating containers and wiring links, so when the pull fails, the actual bridge attachment (D) and the veth link with the renamed `eth1` endpoint (C) are never reached — they were validated at the topology-schema level only. D is correct by construction (name match verified above) and `eth1` follows containerlab's flat-index convention for vrnetlab-based kinds, but neither was exercised end to end; the first run with a real image is where that lands.

**Info:** B/C (`cisco_vios` kind, `type: switch`, flat `ethX` naming, `L2-<date>` tag) are containerlab/vrnetlab-specific and can't be independently verified without the licensed image. Boundary-testing to the pull, plus the free cEOS sibling box covering A+D, is a reasonable stopping point — the README/conf comments now tell a consistent story about the `cisco/vios` build dir and image tag.

mherkazandjian (commented):
Verified all four fixes locally:

- **A (netlab path):** the two new regression tests fail on the pre-fix code (`source_root == ""`) and pass after — reproduced both directions. All four `dirname(config_path)` sites in `manager.py` are now abspath-wrapped, and these are the only two `render_topology` call sites.
- **B (cisco kind/image):** `srl-labs/vrnetlab` indeed has no `cisco/vios_l2` directory (only `cisco/vios`) — the corrected README matches reality. `cisco_vios` + `type: switch` matches containerlab's kind catalog.
- **C (interface):** flat `ethN` is the documented naming for vrnetlab kinds; `Gi0/1` would fail containerlab's endpoint validation.
- **D (shared bridge):** `kind: bridge` + named veth is the canonical way to attach to an existing host bridge, and boxman's netlab layer passes kinds/endpoints through untouched, so no schema conflict on the boxman side.
- Tests: 86/86 netlab tests pass, full suite 1258 passed.

**One request — retarget this PR to `main`.** All four bugs exist identically on `main` (`manager.py:1989`/`2025` and the three boxes on `origin/main` have the same broken code), and the fix has zero dependency on the docker-compose epic: I cherry-picked both commits onto `origin/main` as a dry run and they applied cleanly. As currently targeted, the hybrid boxes stay broken on `main` until the whole epic merges.

(Non-blocking: the `eth1`↔`Gi0/1` mapping is per vrnetlab convention and unverified end-to-end — you flagged this yourself; fine to ship as-is.)

PR #66, opened 2026-07-19
feat(installer): support NixOS, Guix System and Gentoo in the prerequisites checker

mherkazandjian:
## What

Extends the host prerequisites checker (`scripts/installer/check_prerequisites.py`)
to recognise three more Linux distro families — **NixOS**, **Guix System** and
**Gentoo** — alongside the existing Arch / Debian / RHEL support.

`classify_family` now maps `ID=nixos → nixos`, `ID=guix → guix`, `ID=gentoo →
gentoo`, and every downstream consumer (`_CORE_STACK` lookup, `_install_fix`, the
seed/tool-package lookups) routes the new families down their proper per-distro
path instead of the generic "unknown distro" branch.

## Why

Installing the boxman Python package doesn't set up libvirt/KVM. The checker
prints the exact host remediation per distro; previously a NixOS/Guix/Gentoo user
got the "unrecognized distro — install with your package manager" fallback.

## Per-distro model (imperative vs. declarative)

- **Gentoo — imperative (Portage).** Core stack:
  `sudo emerge --ask app-emulation/libvirt app-emulation/qemu
  app-emulation/virt-manager net-dns/dnsmasq app-admin/sudo`
  (`virt-install`/`virt-clone` ship in `app-emulation/virt-manager` — there is no
  standalone `virt-install` atom). `install_cmd` → `sudo emerge --ask <atoms>`;
  per-tool atoms are fully category-qualified. The libvirtd fix keeps the
  `systemctl` command but notes the OpenRC alternative
  (`rc-update add libvirtd default && rc-service libvirtd start`); a code comment
  and the tutorial document the USE-flag caveats
  (`QEMU_SOFTMMU_TARGETS="x86_64"`, libvirt `USE="virt-network qemu"`).
- **NixOS — declarative.** The core stack, the `libvirtd` service and group
  membership are emitted as **advisory** fixes (a `Fix` with no runnable commands,
  so it is printed but never offered to auto-run): a
  `/etc/nixos/configuration.nix` snippet
  (`virtualisation.libvirtd.enable = true;`, `environment.systemPackages`,
  `users.users.<you>.extraGroups = [ "libvirtd" "kvm" ]`) + `sudo nixos-rebuild
  switch`. Individual user CLI tools stay imperative: `install_cmd("nixos", …)` →
  `nix profile install nixpkgs#<pkg>`.
- **Guix System — declarative for services, imperative for user packages.** Core
  stack / service / groups → advisory `/etc/config.scm` snippet (`(service
  libvirt-service-type)`, `supplementary-groups`) + `sudo guix system reconfigure
  /etc/config.scm`. User tools → `install_cmd("guix", …)` → `guix install <pkg>`.

Two presentation helpers (`_show_fix`, the summary's remaining-steps list) now
indent continuation lines of multi-line advisory text. Single-line fixes are
byte-for-byte unchanged, so the **Arch / Debian / RHEL UX is preserved**.

## Tests

`tests/test_check_prerequisites.py` gains: os-release/`classify_family` cases for
the three families; `install_cmd` cases (emerge / `nix profile install
nixpkgs#…` / `guix install`); **table-completeness** checks (every family has a
`_SEED_PKG` and `_TOOL_PKG` column; `_CORE_STACK` holds exactly the imperative
families); and advisory-fix checks (core-stack fix is advisory for nixos/guix,
runnable for the imperative families; new families are never flagged
"unrecognized").

## Testing on hp3 (Rocky Linux 9.6)

All testing ran on the remote host `hp3`, none on the workstation.

- **Unit tests:** `PYTHONPATH=src python3 -m pytest
  tests/test_check_prerequisites.py -v` → **81 passed** (pytest 8.4.2).
- **Regression on hp3's own distro:** `python3 check_prerequisites.py
  --check-only` runs clean; detects `family : rhel`, emits the unchanged
  `sudo dnf install -y …` remediations. (Exit 1 only because hp3's *system*
  python is 3.9 — a real host condition the checker correctly flags, not a
  regression.)
- **Gentoo — fully live in a container:** `docker run --rm -v <checker>:/c:ro
  gentoo/stage3 python3 /c --check-only` → header prints **`family : gentoo`**
  (real `ID=gentoo`, python 3.14 in-container); the emerge remediations render.
  All 12 Gentoo atoms verified against `packages.gentoo.org` (HTTP 200), and no
  standalone `app-emulation/virt-install` (404 → bundled in virt-manager).
- **NixOS / Guix — detection-pipeline validated with injected os-release:** the
  `nixos/nix` image ships **no** `/etc/os-release` and **no** python3 (and there
  is no practical Guix Docker image), so a pure live run isn't possible. Instead
  the authentic NixOS / Guix `os-release` was bind-mounted over `/etc/os-release`
  in a python-capable container, exercising the real `_detect_os → classify_family`
  path: headers print **`family : nixos`** and **`family : guix`**, the core-stack
  fixes render as declarative advisories (no `$`-command), and the seed/tool fixes
  render as `nix profile install nixpkgs#…` / `guix install …`. These two are also
  covered by the fixture unit tests above.

All hp3 validation was read-only (`--check-only` + containers); no guided fix or
package install was run anywhere.

### DIFF
--- a/scripts/installer/check_prerequisites.py
+++ b/scripts/installer/check_prerequisites.py
@@ -90,40 +90,119 @@ def classify_family(osid, id_like):
         return "debian"
     if tokens & {"rhel", "fedora", "centos", "rocky", "almalinux", "ol", "oracle"}:
         return "rhel"
+    if tokens & {"nixos"}:
+        return "nixos"
+    if tokens & {"guix"}:  # ID=guix is always Guix System
+        return "guix"
+    if tokens & {"gentoo"}:
+        return "gentoo"
     return "unknown"
 
 
 # Verbatim per-distro core stack lines (from doc/tutorial/README.md), minus the
 # systemctl/usermod steps which the checker handles as their own fixes.
+#
+# Only *imperative* package-manager distros belong here. NixOS and Guix System
+# are declarative: their libvirt/QEMU stack and the libvirtd service are enabled
+# by editing the system config and rebuilding, never by an imperative "install"
+# command -- see _NIXOS_CORE_ADVICE / _GUIX_CORE_ADVICE, which the core-tools
+# check emits as advisory (printed, never auto-run) fixes instead.
+#
+# Gentoo USE-flag caveats: app-emulation/qemu needs QEMU_SOFTMMU_TARGETS to
+# include x86_64 (e.g. QEMU_SOFTMMU_TARGETS="x86_64") to build qemu-system-x86_64;
+# app-emulation/libvirt needs USE="virt-network qemu" for the QEMU/KVM driver and
+# the default NAT network. virt-install / virt-clone ship in
+# app-emulation/virt-manager. Services are systemd or OpenRC depending on the
+# profile (OpenRC: `rc-update add libvirtd default && rc-service libvirtd start`).
 _CORE_STACK = {
     "arch": "sudo pacman -S --needed libvirt qemu-full virt-install sshpass dnsmasq",
     "debian": ("sudo apt update && sudo apt install -y libvirt-daemon-system "
                "libvirt-clients qemu-kvm virtinst sshpass bridge-utils cloud-image-utils"),
     "rhel": "sudo dnf install -y libvirt qemu-kvm virt-install sshpass genisoimage",
+    "gentoo": ("sudo emerge --ask app-emulation/libvirt app-emulation/qemu "
+               "app-emulation/virt-manager net-dns/dnsmasq app-admin/sudo"),
 }
 
-# Package name for a cloud-init seed-ISO tool, per family.
-_SEED_PKG = {"arch": "libisoburn", "debian": "cloud-image-utils", "rhel": "genisoimage"}
+# Declarative core-stack remediation for NixOS: printed as advisory text, never
+# offered to auto-run (a Fix with no runnable commands).
+_NIXOS_CORE_ADVICE = (
+    "NixOS is declarative -- enable the libvirt/QEMU stack in "
+    "/etc/nixos/configuration.nix:\n"
+    "virtualisation.libvirtd.enable = true;\n"
+    "programs.virt-manager.enable = true;\n"
+    "environment.systemPackages = with pkgs; [ virt-manager qemu cloud-utils ];\n"
+    "users.users.<you>.extraGroups = [ \"libvirtd\" \"kvm\" ];\n"
+    "then apply with: sudo nixos-rebuild switch"
+)
+
+# Declarative core-stack remediation for Guix System (advisory, never auto-run).
+_GUIX_CORE_ADVICE = (
+    "Guix System is declarative -- add libvirt to your operating-system in "
+    "/etc/config.scm:\n"
+    "(use-service-modules virtualization)  ; provides libvirt-service-type\n"
+    "in (services ...): (service libvirt-service-type)\n"
+    "in your (user-account ...): "
+    "(supplementary-groups '(\"libvirt\" \"kvm\" \"wheel\"))\n"
+    "then apply with: sudo guix system reconfigure /etc/config.scm\n"
+    "user CLI tools can also be added with: guix install qemu virt-manager"
+)
+
+# Declarative libvirtd-service remediation (advisory) for NixOS.
+_NIXOS_LIBVIRTD_ADVICE = (
+    "enable libvirtd declaratively: set virtualisation.libvirtd.enable = true; "
+    "in /etc/nixos/configuration.nix, then sudo nixos-rebuild switch"
+)
+
+# Package name for a cloud-init seed-ISO tool, per family. On NixOS/Guix these
+# are individual user CLI tools, so an imperative install is legitimate.
+_SEED_PKG = {
+    "arch": "libisoburn",
+    "debian": "cloud-image-utils",
+    "rhel": "genisoimage",
+    "gentoo": "dev-libs/libisoburn",
+    "nixos": "cloud-utils",
+    "guix": "xorriso",
+}
 
 # Single-package names for individually-missing tools, per family.
 _TOOL_PKG = {
-    "rsync": {"arch": "rsync", "debian": "rsync", "rhel": "rsync"},
-    "sshpass": {"arch": "sshpass", "debian": "sshpass", "rhel": "sshpass"},
-    "ansible": {"arch": "ansible", "debian": "ansible", "rhel": "ansible"},
-    "zstd": {"arch": "zstd", "debian": "zstd", "rhel": "zstd"},
-    "virt-sparsify": {"arch": "guestfs-tools", "debian": "libguestfs-tools", "rhel": "guestfs-tools"},
-    "openssh": {"arch": "openssh", "debian": "openssh-client", "rhel": "openssh-clients"},
+    "rsync": {"arch": "rsync", "debian": "rsync", "rhel": "rsync",
+              "gentoo": "net-misc/rsync", "nixos": "rsync", "guix": "rsync"},
+    "sshpass": {"arch": "sshpass", "debian": "sshpass", "rhel": "sshpass",
+                "gentoo": "net-misc/sshpass", "nixos": "sshpass", "guix": "sshpass"},
+    "ansible": {"arch": "ansible", "debian": "ansible", "rhel": "ansible",
+                "gentoo": "app-admin/ansible", "nixos": "ansible", "guix": "ansible"},
+    "zstd": {"arch": "zstd", "debian": "zstd", "rhel": "zstd",
+             "gentoo": "app-arch/zstd", "nixos": "zstd", "guix": "zstd"},
+    "virt-sparsify": {"arch": "guestfs-tools", "debian": "libguestfs-tools", "rhel": "guestfs-tools",
+                      "gentoo": "app-emulation/libguestfs", "nixos": "guestfs-tools", "guix": "libguestfs"},
+    "openssh": {"arch": "openssh", "debian": "openssh-client", "rhel": "openssh-clients",
+                "gentoo": "net-misc/openssh", "nixos": "openssh", "guix": "openssh"},
 }
 
 
 def install_cmd(family, pkgs):
-    """Build the package-manager install command for `pkgs` (a space string)."""
+    """Build the package-manager install command for `pkgs` (a space string).
+
+    For NixOS and Guix this is only ever used for *individual user CLI tools*
+    (rsync, sshpass, a seed-ISO tool, ...), which may legitimately be installed
+    imperatively into the user profile. The declarative-only bits (the libvirtd
+    service, the core stack, group membership) never go through here -- they are
+    emitted as advisory fixes instead.
+    """
     if family == "arch":
         return "sudo pacman -S --needed " + pkgs
     if family == "debian":
         return "sudo apt update && sudo apt install -y " + pkgs
     if family == "rhel":
         return "sudo dnf install -y " + pkgs
+    if family == "gentoo":
+        return "sudo emerge --ask " + pkgs
+    if family == "nixos":
+        atoms = " ".join("nixpkgs#" + pkg for pkg in pkgs.split())
+        return "nix profile install " + atoms
+    if family == "guix":
+        return "guix install " + pkgs
     return None
 
 
@@ -270,7 +349,12 @@ class Doctor(object):
         return result
 
     def _show_fix(self, fix):
-        print("       " + self.c("fix: " + fix.description, "cyan"))
+        # Descriptions are usually one line; advisory (declarative-distro) fixes
+        # carry a multi-line config snippet, so indent continuation lines.
+        desc_lines = fix.description.splitlines() or [""]
+        print("       " + self.c("fix: " + desc_lines[0], "cyan"))
+        for extra in desc_lines[1:]:
+            print("       " + self.c("     " + extra, "cyan"))
         for cmd in fix.commands:
             print("         " + self.c("$ " + cmd, "dim"))
         if fix.needs_relogin:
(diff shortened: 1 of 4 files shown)


scripts/installer/check_prerequisites.py
```diff
@@ -578,8 +678,17 @@ def libvirtd_service():
             alt = run_capture(["systemctl", "is-active", "virtqemud"])[1].strip()
             if alt == "active":
                 return OK, "virtqemud.service is active (modular libvirt)", None
-            fix = Fix("enable and start libvirtd",
-                      commands=["sudo systemctl enable --now libvirtd"], needs_sudo=True)
+            if self.family == "nixos":
```
mherkazandjian, line 681: Guix System uses Shepherd rather than systemd, so this probe is never registered there because line 694 requires `systemctl`. If the CLI tools are installed but `libvirt-service-type` is absent or inactive, connectivity becomes a blocking FAIL with no fix (`libvirt_conn` explicitly returns `fix=None`). Please add a Guix/Shepherd service probe/advisory and run it without the `systemctl` gate.

scripts/installer/check_prerequisites.py
```diff
@@ -90,40 +90,119 @@ def classify_family(osid, id_like):
...
-_SEED_PKG = {"arch": "libisoburn", "debian": "cloud-image-utils", "rhel": "genisoimage"}
+# Declarative core-stack remediation for NixOS: printed as advisory text, never
+# offered to auto-run (a Fix with no runnable commands).
+_NIXOS_CORE_ADVICE = (
+    "NixOS is declarative -- enable the libvirt/QEMU stack in "
+    "/etc/nixos/configuration.nix:\n"
+    "virtualisation.libvirtd.enable = true;\n"
+    "programs.virt-manager.enable = true;\n"
+    "environment.systemPackages = with pkgs; [ virt-manager qemu cloud-utils ];\n"
+    "users.users.<you>.extraGroups = [ \"libvirtd\" \"kvm\" ];\n"
```
mherkazandjian, line 134: The earlier `/dev/kvm access` probe still offers runnable `sudo usermod -aG kvm $USER` for every family (lines 616–618), before this declarative group advice is reached. Under `--yes` that mutates NixOS/Guix imperatively and will be overwritten by a rebuild/reconfigure. Please make that KVM-access remediation advisory for these two families as well.

scripts/installer/check_prerequisites.py
```diff
@@ -90,40 +90,119 @@ def classify_family(osid, id_like):
...
+    emitted as advisory fixes instead.
+    """
     if family == "arch":
         return "sudo pacman -S --needed " + pkgs
     if family == "debian":
         return "sudo apt update && sudo apt install -y " + pkgs
     if family == "rhel":
         return "sudo dnf install -y " + pkgs
+    if family == "gentoo":
+        return "sudo emerge --ask " + pkgs
```
mherkazandjian, line 200: When `_check_sudo_rights` detects that sudo is missing, it calls this path with bare `sudo`, producing `sudo emerge --ask sudo`. That cannot run without sudo, and Gentoo requires the qualified atom `app-admin/sudo` (already used in `_CORE_STACK`). Please special-case the sudo bootstrap as a non-runnable/root advisory, or supply a root-capable command with the qualified atom.

scripts/installer/check_prerequisites.py
```diff
@@ -90,40 +90,119 @@ def classify_family(osid, id_like):
...
+    "rsync": {"arch": "rsync", "debian": "rsync", "rhel": "rsync",
+              "gentoo": "net-misc/rsync", "nixos": "rsync", "guix": "rsync"},
+    "sshpass": {"arch": "sshpass", "debian": "sshpass", "rhel": "sshpass",
+                "gentoo": "net-misc/sshpass", "nixos": "sshpass", "guix": "sshpass"},
+    "ansible": {"arch": "ansible", "debian": "ansible", "rhel": "ansible",
+                "gentoo": "app-admin/ansible", "nixos": "ansible", "guix": "ansible"},
+    "zstd": {"arch": "zstd", "debian": "zstd", "rhel": "zstd",
+             "gentoo": "app-arch/zstd", "nixos": "zstd", "guix": "zstd"},
+    "virt-sparsify": {"arch": "guestfs-tools", "debian": "libguestfs-tools", "rhel": "guestfs-tools",
+                      "gentoo": "app-emulation/libguestfs", "nixos": "guestfs-tools", "guix": "libguestfs"},
```
mherkazandjian, line 178: **[should-fix] Both new `virt-sparsify` cells point at packages that do not ship `virt-sparsify`.**

Upstream split the `virt-*` tools out of libguestfs into the separate **guestfs-tools** project at 1.46 (2021), and both of these new cells predate-the-split in spirit:

- **Gentoo**: the tree has `app-emulation/guestfs-tools` (1.52.3/1.54.0/1.55.6) alongside `app-emulation/libguestfs` (1.54.1-1.58.1). `libguestfs-1.58.1.ebuild` contains zero mentions of sparsify, while `guestfs-tools`' own `pkg_postinst` names `virt-sparsify` among its tools (built with the default-on `+ocaml` flag). A user who accepts this fix emerges the library and the `virt-sparsify` WARN persists. Fix: `"gentoo": "app-emulation/guestfs-tools"`.
- **Guix**: `libguestfs` in Guix is 1.56.2 - post-split, so no `virt-sparsify` - and Guix packages no `guestfs-tools` at all (packages.guix.gnu.org search and `gnu/packages/virtualization.scm` both come up empty). `guix install libguestfs` succeeds yet can never yield the binary. Suggest dropping the guix cell and special-casing an honest advisory ("virt-sparsify is not packaged in Guix; build guestfs-tools from source or skip `boxman storage compact`"), rather than printing a fix that cannot work.

The other four cells are fine (arch/rhel `guestfs-tools`, debian `libguestfs-tools` -> transitional to guestfs-tools, nixos `guestfs-tools` exists in pkgs/by-name).

scripts/installer/check_prerequisites.py
```diff
@@ -578,8 +678,17 @@ def libvirtd_service():
...
-                      commands=["sudo systemctl enable --now libvirtd"], needs_sudo=True)
+            if self.family == "nixos":
+                fix = Fix(_NIXOS_LIBVIRTD_ADVICE, commands=[])  # declarative: no auto-run
+            elif self.family == "gentoo":
+                # systemd profile: systemctl works. OpenRC profile: use rc-service.
+                fix = Fix("enable and start libvirtd (systemd profile; on an OpenRC "
+                          "profile use `sudo rc-update add libvirtd default && "
+                          "sudo rc-service libvirtd start`)",
+                          commands=["sudo systemctl enable --now libvirtd"], needs_sudo=True)
+            else:
```
mherkazandjian, line 689: **[should-fix] Guix System lands in the systemd `else` branch.**

Guix System boots GNU Shepherd, so `sudo systemctl enable --now libvirtd` can never be the right remediation for `family == "guix"`. Today this is masked: the whole check is gated on `have("systemctl")` a few lines below, and Guix System ships no `systemctl`, so the check SKIPs before this fix is ever built. But the wrong branch is latent (a systemctl shim in a container image, or any future de-gating of this check, surfaces it), and it is inconsistent with the rest of this PR - `groups()` and `_core_stack_fix()` both special-case guix, this dispatch does not.

Suggest mirroring the NixOS treatment with a `_GUIX_LIBVIRTD_ADVICE` advisory:

```python
elif self.family == "guix":
    fix = Fix("enable libvirtd declaratively: add (service libvirt-service-type) "
              "to /etc/config.scm, then sudo guix system reconfigure /etc/config.scm "
              "(start immediately with: sudo herd start libvirtd)", commands=[])
```

Also: none of the new unit tests pin this per-family dispatch (they cover `_core_stack_fix`/`_install_fix` only) - a small table test over `libvirtd_service`'s fix construction would have caught this.

scripts/installer/check_prerequisites.py
```diff
@@ -600,13 +709,27 @@ def libvirt_conn():
...
+            # NixOS names the libvirt group `libvirtd`; everyone else `libvirt`.
+            libvirt_group = "libvirtd" if self.family == "nixos" else "libvirt"
+            want = [libvirt_group, "kvm"]
             missing = [g for g in want if g not in names]
             if not missing:
                 return OK, "user '%s' is in: %s" % (user, ", ".join(want)), None
-            fix = Fix("add yourself to the libvirt and kvm groups",
-                      commands=["sudo usermod -aG libvirt,kvm $USER"],
-                      needs_sudo=True, needs_relogin=True)
+            if self.family == "nixos":
```
mherkazandjian, line 718: **[should-fix] The declarative-groups design misses the second `usermod` path.**

This hunk correctly makes group membership advisory on NixOS/Guix - but the `/dev/kvm access` check (`kvm_access`, ~line 616, untouched by this PR) still emits a runnable `Fix(commands=["sudo usermod -aG kvm $USER"])` for **every** family. Because that fix is runnable, `--yes` auto-executes it on NixOS/Guix, exactly what this PR's model says must not happen for group membership on declarative distros.

Concrete failure: on Guix System the `/etc/group` edit is silently reverted by the next `guix system reconfigure`; on NixOS it bypasses `configuration.nix` (and with `users.mutableUsers = false` is discarded on the next rebuild) - in both cases the user loses `/dev/kvm` access again later with no hint why.

Suggest routing `kvm_access`'s fix through the same family dispatch as here (advisory `extraGroups = [ "kvm" ]` / `supplementary-groups` text with `commands=[]` for nixos/guix), or extracting a shared group-fix helper so the two checks cannot drift.

scripts/installer/check_prerequisites.py
```diff
@@ -90,40 +90,119 @@ def classify_family(osid, id_like):
...
     if family == "arch":
         return "sudo pacman -S --needed " + pkgs
     if family == "debian":
         return "sudo apt update && sudo apt install -y " + pkgs
     if family == "rhel":
         return "sudo dnf install -y " + pkgs
+    if family == "gentoo":
+        return "sudo emerge --ask " + pkgs
+    if family == "nixos":
+        atoms = " ".join("nixpkgs#" + pkg for pkg in pkgs.split())
```
mherkazandjian, line 202: **[nit] `nix profile install nixpkgs#...` presumes flakes + nix-command are enabled.**

Stock Nix/NixOS still ships with the `nix-command` and `flakes` experimental features disabled, so on a default NixOS install every one of these suggested commands fails with `error: experimental Nix feature 'nix-command' is disabled`. Most NixOS users have them enabled, hence nit - but since this checker exists precisely for first-time setup, consider a parenthetical in the advice (`nix.settings.experimental-features = [ "nix-command" "flakes" ];`) or the classic fallback `nix-env -iA nixos.<pkg>`.

scripts/installer/check_prerequisites.py
```diff
@@ -600,13 +709,27 @@ def libvirt_conn():
...
-                      commands=["sudo usermod -aG libvirt,kvm $USER"],
-                      needs_sudo=True, needs_relogin=True)
+            if self.family == "nixos":
+                fix = Fix("add your user to the libvirtd/kvm groups declaratively: "
+                          "users.users.<you>.extraGroups = [ \"libvirtd\" \"kvm\" ]; "
+                          "in /etc/nixos/configuration.nix, then sudo nixos-rebuild switch",
+                          commands=[], needs_relogin=True)
+            elif self.family == "guix":
+                fix = Fix("add your user to the libvirt/kvm groups declaratively: put "
+                          "(supplementary-groups '(\"libvirt\" \"kvm\")) in your "
```
mherkazandjian, line 725: **[nit] Replace-semantics footgun in the copy-paste advice.**

`supplementary-groups` is a single field of `user-account` whose typical value is `'("wheel" "netdev" "audio" "video")`. Told to "put (supplementary-groups '(\"libvirt\" \"kvm\")) in your user-account", a literal copy-paster *replaces* that field and drops `wheel` - losing sudo membership at the next `guix system reconfigure`. Phrasing it as "add \"libvirt\" and \"kvm\" to your user-account's existing supplementary-groups" avoids the trap. Same nuance in `_GUIX_CORE_ADVICE` (line 145 keeps `wheel` but drops `netdev`/`audio`/`video`) and the tutorial's Guix block (doc/tutorial/README.md).

mherkazandjian (commented):
Deep review of the NixOS / Guix System / Gentoo extension, with the fix-runner machinery and every new package mapping checked against the real distro repos.

**Verified clean** (no action needed):
- **Advisory-fix safety:** a `Fix` with `commands=[]` (`runnable == False`) short-circuits in `_handle_fix` *before* the `--yes`/prompt path, so declarative advisories are printed and summarized but never offered with "run the above now?" and never auto-run under `--yes`. The multi-line advisory rendering is correct in both `_show_fix` and `_summary`.
- **`_core_stack_fix()` extraction:** behavior-identical to the old inline logic for arch/debian/rhel and the unknown-distro fallback (same description, commands, flags) - no regression.
- **NixOS advisory completeness:** `virtualisation.libvirtd.enable = true` adds `cfg.package` + `cfg.qemu.package` to `environment.systemPackages` (nixos/modules/virtualisation/libvirtd.nix), i.e. `virsh`, `qemu-img`, `qemu-system-x86_64`; nixpkgs `virt-manager` ships `virt-install`/`virt-clone`; nixpkgs `cloud-utils` 0.33 ships `cloud-localds` (wrapped with cdrkit on PATH, so it actually works). All five required core binaries are covered.
- **Guix advisory completeness:** `libvirt-service-type` extends `profile-service-type` with libvirt + qemu (gnu/services/virtualization.scm), and Guix `virt-manager` 5.0.0 ships `virt-install`/`virt-clone` - so the advisory (service + `guix install qemu virt-manager`) covers all five binaries. `xorriso`/`sshpass` package names are valid.
- **Gentoo USE-flag guidance:** matches the ebuilds (`+qemu`/`+virt-network` are libvirt defaults; the caveat is belt-and-braces and correct). `dev-libs/libisoburn` is the right xorriso atom; `app-emulation/virt-manager` does ship `virt-install`/`virt-clone`.
- **`classify_family`:** ordering is safe; `ID=guix` => Guix System is sound; bonus - Funtoo/Calculate (`ID_LIKE=gentoo`) now get correct emerge advice instead of the unknown fallback.
- **Style/compat:** all new code stays stdlib-only, `%`-formatting, Python-3.6-safe.

**No blocking findings.** Ranked:

*Should-fix*
1. `_TOOL_PKG["virt-sparsify"]` - both **new** cells name packages that do not ship `virt-sparsify`: Gentoo's lives in `app-emulation/guestfs-tools` (not `libguestfs`), and Guix does not package it at all (its `libguestfs` is 1.56.2, post-upstream-split). Following the printed fix installs something and the WARN persists. (check_prerequisites.py:178)
2. Guix System falls into the systemd `else` of `libvirtd_service` -> `sudo systemctl enable --now libvirtd`, wrong for Shepherd. Latent today only because the check is gated on `have("systemctl")`, and inconsistent with `groups()`/`_core_stack_fix()` which do special-case guix. (check_prerequisites.py:689)
3. The untouched `/dev/kvm access` check still suggests - and under `--yes` **auto-runs** - `sudo usermod -aG kvm $USER` on NixOS/Guix, contradicting this PR's own declarative-groups model. (noted at check_prerequisites.py:718)

*Nits*
4. `nix profile install nixpkgs#...` requires the still-off-by-default `nix-command`/`flakes` experimental features on stock Nix/NixOS. (check_prerequisites.py:202)
5. The Guix groups advice invites replacing `supplementary-groups` verbatim, which drops `wheel` (sudo) for a literal copy-paster. (check_prerequisites.py:725)

Minor test note: the new dispatch tests are good, but nothing pins the `libvirtd_service` per-family dispatch (a small table test would have caught finding 2), nor the multi-line renderers. And a lesson the virt-sparsify row teaches for future columns: validating atoms by HTTP 200 checks package *existence*, not that the package ships the binary the checker probes for.

PR #70, opened 2026-07-23
feat(docker-compose): Phase 6 CLI/UX parity for containers (#54)

Orski174:
Phase 6 of the docker-compose provider epic (#42). Closes #54.

## What this does
Makes every boxman verb treat a mixed libvirt + docker-compose project
uniformly — containers show up alongside VMs and no verb tracebacks on an
unsupported op.

## New verb: `boxman exec`
`boxman exec <cluster>.<box>` runs `docker compose exec` on a container: an
interactive shell (default `sh`, `--shell bash`) or a one-shot command after
`--` (`boxman exec services.cache -- redis-cli ping`). A bare `<box>` works
when unique across dc clusters. `boxman ssh` stays VM-only.

> **Revises decision D2** (which had `ssh` transparently docker-exec): a
> distinct `exec` verb reads clearer than overloading `ssh`. D2 + AC17 updated
> in `design.md`.

## Verb parity
- **`ps`** — containers listed with a `Provider` column + live `State`/health;
  a dc-only project skips the virsh query.
- **`connect_info`** — per-container section: status, published ports, the
  `boxman exec` entry point.
- **Inventory / `run` / `tasks`** — containers become Ansible hosts with
  `ansible_connection: community.docker.docker` + `ansible_host:
  <project>-<box>-1`, grouped per cluster; `env.sh`/`ansible.cfg` now generated
  for dc-only projects. So `boxman run --cmd '…'` and `tasks:` hit containers
  and VMs alike. **Prereq:** the `community.docker` collection + Docker SDK on
  the control host (documented).
- **`control`** — `suspend`→`pause`, `resume`→`unpause`, `start`→`start`;
  `save` logs a clean "unsupported" message and skips dc clusters.

## Design call: `list`
`boxman list` stays a **config-free cross-project registry** view (loading
every registered project's config there would add fragility). Per-cluster
provider + container status for the current project is surfaced by
`ps`/`connect_info`. Happy to add a cached provider column if you'd prefer it
in `list`.

## Testing
- **Host unit:** 1295 passed (+19: exec command build + target resolution, ps
  container rows, control dc branch, `container_status` parsing, pause/unpause,
  inventory `community.docker` entries). VM inventory/ps output preserved.
- **Live e2e in stage01** (dc-only, pure docker):
  - **AC16** `boxman ps` shows containers with provider + running/healthy state
  - **AC17** `boxman exec services.cache -- redis-cli ping` → `PONG`; `nginx -v`
  - **AC18** `ansible all -m raw` reaches **both** containers via the
    boxman-generated inventory + `community.docker`
  - `control suspend`→paused / `resume`→running / `save`→clean message

### DIFF
--- a/src/boxman/manager.py
+++ b/src/boxman/manager.py
@@ -284,14 +284,24 @@ class BoxmanManager:
         }
 
     def _compose_project_for(self, cluster_name: str) -> str:
-        """The ``docker compose -p`` project name for a dc cluster — mirrors
-        :meth:`DockerComposeSession._compose_project` so container names can be
-        derived (``<project>-<box>-1``) at inventory-render time without a live
-        session or a docker query."""
-        from boxman.providers.docker_compose.session import _sanitize_project_name
-        dc = (self.config.get('provider') or {}).get('docker-compose') or {}
-        base = dc.get('project_name') or self.config.get('project') or 'boxman'
-        return _sanitize_project_name(f"{base}_{cluster_name}")
+        """The ``docker compose -p`` project name for a dc cluster, so container
+        names can be derived (``<project>-<box>-1``) at inventory-render time
+        without a live session or docker query. Single-sourced with the
+        session's ``_compose_project`` so the inventory ``ansible_host`` can
+        never diverge from the real compose container name."""
+        from boxman.providers.docker_compose.session import compose_project_name
+        return compose_project_name(self.config, cluster_name)
+
+    def _select_dc_clusters(self, cli_args) -> list[tuple[str, dict]]:
+        """docker-compose clusters selected by ``--cluster`` (an unset/absent
+        ``--cluster`` selects all). ``(name, cfg)`` pairs — empty for a
+        libvirt-only project."""
+        wanted = getattr(cli_args, 'cluster', None)
+        return [
+            (name, cluster)
+            for name, cluster in self._compose_clusters.items()
+            if wanted in (None, name)
+        ]
 
     def _vm_cluster_map(self) -> dict[str, str]:
         """
@@ -5143,7 +5153,7 @@ class BoxmanManager:
         for vm_name, _ in cls.process_vm_list(cli_args):
             cls.session_for_vm(vm_name).suspend_vm(vm_name)
             cls.logger.info(f"vm {vm_name} suspended")
-        for cluster_name, cluster in cls._compose_clusters.items():
+        for cluster_name, cluster in cls._select_dc_clusters(cli_args):
             cls._dc_session(cluster_name).pause_cluster(cluster_name, cluster)
 
     @staticmethod
@@ -5155,7 +5165,7 @@ class BoxmanManager:
         for vm_name, _ in cls.process_vm_list(cli_args):
             cls.session_for_vm(vm_name).resume_vm(vm_name)
             cls.logger.info(f"VM {vm_name} resumed")
-        for cluster_name, cluster in cls._compose_clusters.items():
+        for cluster_name, cluster in cls._select_dc_clusters(cli_args):
             cls._dc_session(cluster_name).unpause_cluster(cluster_name, cluster)
 
     @staticmethod
@@ -5167,7 +5177,7 @@ class BoxmanManager:
         """
         for vm_name, workdir in cls.process_vm_list(cli_args):
             cls.session_for_vm(vm_name).save_vm(vm_name, workdir)
-        for cluster_name in cls._compose_clusters:
+        for cluster_name, _cluster in cls._select_dc_clusters(cli_args):
             cls.logger.warning(
                 f"'control save' is not supported for docker-compose cluster "
                 f"'{cluster_name}' — containers have no save-to-file state; use "
@@ -5185,7 +5195,7 @@ class BoxmanManager:
                 cls.session_for_vm(vm_name).restore_vm(vm_name, workdir)
             else:
                 cls.session_for_vm(vm_name).start_vm(vm_name)
-        for cluster_name, cluster in cls._compose_clusters.items():
+        for cluster_name, cluster in cls._select_dc_clusters(cli_args):
             if getattr(cli_args, "restore", False):
                 cls.logger.info(
                     f"[{cluster_name}] --restore has no docker-compose "
@@ -5433,7 +5443,7 @@ class BoxmanManager:
         records = []
         for idx, (cluster_name, vm_name, full_name) in enumerate(vm_list):
             virsh_id, state = vm_info.get(full_name, ("-", "not created"))
-            rec = {"id": idx, "cluster": cluster_name, "name": vm_name,
+            rec = {"id": idx, "cluster": cluster_name, "vm": vm_name,
                    "provider": cls.provider_type_for_cluster(cluster_name),
                    "state": state}
             if provider_info:
@@ -5458,7 +5468,7 @@ class BoxmanManager:
                 if row:
                     state = row["state"] + (
                         f" ({row['health']})" if row.get("health") else "")
-                rec = {"id": "-", "cluster": cluster_name, "name": box_name,
+                rec = {"id": "-", "cluster": cluster_name, "vm": box_name,
                        "provider": "docker-compose", "state": state}
                 if provider_info:
                     rec["virsh_id"] = "-"
@@ -5477,11 +5487,11 @@ class BoxmanManager:
         if provider_info:
             headers = ("Id", "Cluster", "Name", "Provider", "State",
                        "Virsh Id", "Virsh Name")
-            rows = [(str(r["id"]), r["cluster"], r["name"], r["provider"],
+            rows = [(str(r["id"]), r["cluster"], r["vm"], r["provider"],
                      r["state"], r["virsh_id"], r["virsh_name"]) for r in records]
         else:
             headers = ("Id", "Cluster", "Name", "Provider", "State")
-            rows = [(str(r["id"]), r["cluster"], r["name"], r["provider"],
+            rows = [(str(r["id"]), r["cluster"], r["vm"], r["provider"],
                      r["state"]) for r in records]
 
         col_count = len(headers)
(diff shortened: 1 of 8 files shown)


src/boxman/scripts/cli_parser.py
```diff
@@ -1077,6 +1077,43 @@ def parse_args():
...
+    )
+    parser_exec.add_argument(
+        '--shell',
+        type=str,
+        default=None,
+        help='interactive shell to open when no command is given (default: sh)'
+    )
+    parser_exec.add_argument(
+        'cmd',
+        nargs=argparse.REMAINDER,
```
mherkazandjian: Because `cmd` uses `argparse.REMAINDER`, options after `target` are swallowed into `cmd`. The documented `boxman exec services.web --shell bash` parses as `shell=None, cmd=['--shell', 'bash']` and tries to execute `--shell` in the container. Please change the parsing strategy (or preprocess the `--` command separator) so `--shell` works in the advertised position, and add an end-to-end parser test.

src/boxman/manager.py
```diff
@@ -5063,39 +5137,60 @@ def process_vm_list(self, cli_args):
...
         """
-        Suspend (pause) the VMs in the cluster.
+        Suspend the machines: libvirt VMs → virsh suspend; docker-compose
+        containers → ``docker compose pause``.
         """
         for vm_name, _ in cls.process_vm_list(cli_args):
             cls.session_for_vm(vm_name).suspend_vm(vm_name)
             cls.logger.info(f"vm {vm_name} suspended")
+        for cluster_name, cluster in cls._compose_clusters.items():
+            cls._dc_session(cluster_name).pause_cluster(cluster_name, cluster)
```
mherkazandjian, line 5157: This ignores `cli_args.vms` and pauses every service in every docker-compose cluster. For example, `boxman control suspend --vms web` will also pause `cache` (and services in other dc clusters), despite the command's targeting option. The same issue applies to resume/start below. Please resolve the selected names into per-cluster container lists and pass them to `pause_cluster`/`unpause_cluster`/`start_cluster`, skipping unselected clusters; add mixed-project coverage.

src/boxman/scripts/cli_parser.py
mherkazandjian: **Bug (fix before merge): `--shell` is silently ignored whenever it appears after the target — which is how the example on line 1094 writes it.**

`nargs=argparse.REMAINDER` makes `cmd` swallow *everything* after the first positional, options included. Re-ran this exact parser as `_main()` does (`parse_known_args`, Python 3.12.3), then applied `exec_container`'s post-processing:

```
['exec','services.web','--shell','bash']              -> shell=None   cmd=['--shell','bash']
    runs: docker compose -p demo_services exec -T web --shell bash
['exec','services.web','--shell','bash','--','redis-cli','ping']
                                                      -> shell=None   cmd=['--shell','bash','--','redis-cli','ping']
['exec','--shell','bash','services.web']              -> shell='bash' cmd=[]    # only placement that works
['exec','services.web','--','redis-cli','ping']       -> shell=None   cmd=['redis-cli','ping']   # OK
```

So the help text's own example (`$ boxman exec services.web --shell bash`) does **not** open a bash shell: `shell` stays `sh`, the exec turns non-interactive (`-T`), and `--shell bash` is passed as the in-container command — `docker compose exec` treats everything after the service as the command, so it tries to run a program named `--shell` and fails. `exec_container` can't rescue it either: it only strips a *leading* literal `--`, and in the second case above the interior `--` is forwarded into the container command too.

Suggested fixes (either works):
- Drop `REMAINDER`: split the argv on the first `--` *before* argparse sees it (everything after `--` is `cmd`), so `--shell` may appear anywhere before the separator; or
- Keep `REMAINDER`, but (a) fix the example to `boxman exec --shell bash services.web`, and (b) fail loudly in `exec_container` when `cmd[0]` starts with `-` — a boxman flag should never be silently forwarded into the container.

Either way, please add a parser-level regression test: the new Phase-6 tests construct `SimpleNamespace(target=…, cmd=[], shell=None)` directly, which is exactly why argparse behavior went untested.

src/boxman/manager.py
```diff
@@ -5063,39 +5137,60 @@ def process_vm_list(self, cli_args):
...
     def suspend_vm(cls, cli_args):
         """
-        Suspend (pause) the VMs in the cluster.
+        Suspend the machines: libvirt VMs → virsh suspend; docker-compose
+        containers → ``docker compose pause``.
         """
         for vm_name, _ in cls.process_vm_list(cli_args):
             cls.session_for_vm(vm_name).suspend_vm(vm_name)
             cls.logger.info(f"vm {vm_name} suspended")
+        for cluster_name, cluster in cls._compose_clusters.items():
```
mherkazandjian: **All four `control` dc branches act on *every* dc cluster in the project, unconditionally** (same pattern here and in `resume_vm`, `save_vm`, `start_vm`). There is no way to scope them: the `control` subcommands only define `--vms` (no `--cluster`), and `--vms` has no docker-compose meaning — so `boxman control suspend --vms=node01` on a mixed project pauses every container in every dc cluster.

Two consistency points:
- This same PR added `boxes=` scoping to `pause_cluster`/`unpause_cluster` in the session layer, but the manager never passes it — that plumbing is dead on arrival.
- The phase-7 branch scopes its dc snapshot ops via `_select_dc_clusters(cli_args)` honoring `--cluster`; once that lands, `control` will be the odd verb out.

(For fairness: the VM path doesn't honor `--vms` either — `process_vm_list` ignores `cli_args`, and that is pre-existing on `main`, not this PR's doing. But these loops add a second, larger unscoped blast radius.)

Suggestion: reuse the phase-7-style selection here (support `--cluster` on the control subcommands), and/or map `--vms` entries of the form `<cluster>.<box>` onto the session's `boxes=` parameter.

src/boxman/manager.py
```diff
@@ -5304,58 +5399,90 @@ def ps(cls, cli_args):
...
+                        virsh_id, virsh_name, state = parts[0], parts[1], parts[2].strip()
+                        vm_info[virsh_name] = (virsh_id, state)
 
-        # Build records
+        # Build records: libvirt VMs first (numeric ids), then dc containers.
         records = []
         for idx, (cluster_name, vm_name, full_name) in enumerate(vm_list):
             virsh_id, state = vm_info.get(full_name, ("-", "not created"))
-            rec = {"id": idx, "cluster": cluster_name, "vm": vm_name, "state": state}
+            rec = {"id": idx, "cluster": cluster_name, "name": vm_name,
```
mherkazandjian: Compat note (intentional, I assume, but worth stating): `ps --json` records rename `vm` → `name` and add `provider` — on `main` today the key is `"vm"`, so any consumer doing e.g. `jq '.[].vm'` breaks. Container rows additionally emit `"id": "-"` (a string) alongside the integer ids of VM rows, so `id` is no longer uniformly numeric either. Fine as a deliberate generalization — it just deserves a changelog/release-note line, and maybe a decision on whether to keep `vm` as a deprecated alias for one release.

src/boxman/manager.py
```diff
@@ -265,6 +283,16 @@ def _compose_clusters(self) -> dict[str, Any]:
             if self._is_compose_cluster(name)
         }
 
+    def _compose_project_for(self, cluster_name: str) -> str:
```
mherkazandjian, line 286: Nit (future-drift risk, not a bug today): this duplicates `DockerComposeSession._compose_project`. I verified they cannot diverge *currently* — dc sessions are constructed from the raw project config in every path (`app.py` uses `session_config = manager.config` for non-libvirt providers, `_dc_session` likewise passes `self.config`, and `update_provider_config_with_runtime` is a dc no-op) — so the inventory's `ansible_host` does match the real compose container name.

But this is exactly the derivation Ansible connectivity hangs on: if either copy ever learns a new source (say, app-level `providers.docker-compose.project_name`, or a per-cluster override), the other silently diverges and `community.docker` stops resolving containers. Suggest single-sourcing it — e.g. a module-level `compose_project_name(config, cluster_name)` that both call, or at minimum a unit test asserting `manager._compose_project_for(c) == session._compose_project(c)`.

mherkazandjian (commented):
## Review — Phase 6 CLI/UX parity

Strong, careful PR overall: the inventory generalization (`host_extra_vars`), the never-raise status probes in `ps`/`connect_info`, the `-T`/shlex-quoting discipline in `exec_command`, and the `save` warn+skip behavior are all solid, and the 19 new unit tests pass locally (Python 3.12.3). One real bug should be fixed before merge, plus a scoping gap and two smaller notes — details on the inline comments.

**Bug — fix before merge: `boxman exec <target> --shell <x>` silently ignores `--shell`.** The `cmd` positional uses `nargs=argparse.REMAINDER`, which swallows any option appearing after the target — including the exec help text's *own example*. I re-ran this branch's real parser exactly as `_main()` does (`parse_known_args`, Python 3.12.3) and then applied `exec_container`'s post-processing:

```
['exec','services.web','--shell','bash']              -> shell=None   cmd=['--shell','bash']
    runs: docker compose -p demo_services exec -T web --shell bash        # fails in-container
['exec','services.web','--shell','bash','--','redis-cli','ping']
                                                      -> shell=None   cmd=['--shell','bash','--','redis-cli','ping']
['exec','--shell','bash','services.web']              -> shell='bash' cmd=[]    # only placement that works
['exec','services.web','--','redis-cli','ping']       -> shell=None   cmd=['redis-cli','ping']   # OK
['exec','services.web']                               -> shell=None   cmd=[]                     # OK
```

So `boxman exec services.web --shell bash` does **not** open a bash shell: `shell` stays `sh`, the exec goes non-interactive (`-T`), and `--shell bash` becomes the in-container command, which fails. Suggested fixes on the inline comment in `cli_parser.py`.

**Medium — `control` dc branches act on every dc cluster unconditionally.** No `--cluster` flag exists on the `control` subcommands, `--vms` has no dc meaning, and the `boxes=` scoping that this same PR added to `pause_cluster`/`unpause_cluster` is never passed by the manager. The phase-7 branch already scopes its dc snapshot ops via `_select_dc_clusters(cli_args)` — `control` should get the same treatment. Details on the inline comment in `suspend_vm`.

**Notes (low):**
- `ps --json` schema change: `vm` → `name`, new `provider` key, and container rows use string `"-"` ids — breaks consumers reading `.vm` (on `main` today the key is `"vm"`). Worth a changelog line.
- `_compose_project_for` duplicates the session's `_compose_project` derivation. Verified equivalent today in all paths, but it is the derivation `ansible_host` hangs on — see inline for a single-sourcing suggestion.

**Verified while reviewing (no action needed):**
- `provider_type_for_cluster` exists (manager.py:161) — the `ps` VM branch is correct.
- Inventory `ansible_host = <project>-<box>-1` matches the real compose container name in all current paths: dc sessions are constructed from the raw project config both in the normal path (`app.py`: `session_config = manager.config` for non-libvirt) and in `_dc_session`, so the manager/session project-name derivations cannot diverge today.
- `container_status`/`_format_ports` correctly handle NDJSON-or-array output and never raise; `exec_command` uses `-T` for one-shots and shlex-quotes everything.
- The `exec` dispatch insertion in `app.py` (an `if` block before the pre-existing `else`) is behavior-preserving since every earlier branch exits; the light `exec`/`ps` dispatch paths do get workspace-default workdirs resolved (`BoxmanManager.__init__` → `resolve_workspace_defaults`), and a `ConfigError` from exec surfaces as a clean exit-2 via `main()`'s `BoxmanError` handler.
- `process_vm_list` ignoring `--vms` is pre-existing (identical on `main`), not introduced here — mentioned only as context on the control comment.

Verdict: nice work — please fix the `--shell` parsing before merge; the control-scoping point is worth deciding on now (even if the answer is a follow-up), and the two notes are take-or-leave.

PR #71, opened 2026-07-23
feat(docker-compose): Phase 7 container snapshots (docker commit, #55)

Orski174:
Phase 7 of the docker-compose provider epic (#42). Closes #55.
Base: `feat/docker-compose-provider` (parallel to Phase 6 per the dep chain; I'll rebase onto updated feat if #70 merges first).

## What this does
Implements the Phase-0 decision **D3**: `docker commit`-backed snapshots for
docker-compose clusters, so `boxman snapshot take/restore/list/delete` work on
containers and never traceback on a mixed project.

## Behaviour
- **`take --name <n>`** → `docker commit` each container to
  `boxman/<project>_<box>:<n>` + `snapshots.json` in the cluster workdir.
- **`restore --name <n>`** → regenerate the compose file with the snapshot
  image tags + `up --force-recreate` (a later `boxman up` reverts to the
  declared images — a point-in-time recreate, not a permanent pin).
- **`list` / `delete --name <n>`** → metadata + `docker image rm`.
- Honours `--cluster`; `restore` resolves the latest snapshot when `--name` is
  omitted.

> **Named volumes are NOT captured** — `docker commit` snapshots only a
> container's writable layer, not volume data. This is the key divergence from
> libvirt external snapshots, warned on every `take` and documented.

## Container state across down/up
Documented (independent of snapshots): `down` keeps named volumes (data
survives down/up), the container filesystem is rebuilt from the declared image
on `up`, `destroy` removes volumes.

## Testing
- **Host unit:** 1290 passed (+18: commit/image_rm/force-recreate command
  strings, take commits + writes metadata, no-containers raises, list/delete
  round-trip, restore regenerates with snapshot images, tag sanitization,
  manager dc-branch routing).
- **Live e2e in stage01** (dc-only, pure docker): `snapshot take v1` →
  write "changed" into the container FS → `snapshot restore v1` → **the
  filesystem rolls back to the snapshot state**; `list` shows it; `delete`
  removes the image + metadata; the volume-not-captured warning fires.


src/boxman/providers/docker_compose/session.py
```diff
@@ -131,6 +134,172 @@ def destroy_cluster(self, cluster_name: str, cluster_cfg: dict[str, Any]) -> boo
...
+        for box in (cluster_cfg.get("boxes") or {}):
+            container = names.get(box)
+            if not container:
+                self.logger.warning(
+                    f"[{cluster_name}] box '{box}' has no container — skipping "
+                    f"in snapshot '{snapshot_name}'."
+                )
+                continue
+            tag = _snapshot_tag(project, box, snapshot_name)
+            result = runner.commit(container, tag)
```
mherkazandjian, line 289: A repeated snapshot name is unsafe here: each successful `docker commit` immediately retargets the existing tag, but metadata is only updated after all boxes finish. If a later commit fails, the previously recorded snapshot becomes a mix of old and new box states. Please reject an existing snapshot name up front, or commit to temporary unique tags and only publish/replace them after every commit succeeds.

src/boxman/providers/docker_compose/session.py
```diff
@@ -131,6 +134,172 @@ def destroy_cluster(self, cluster_name: str, cluster_cfg: dict[str, Any]) -> boo
...
+        meta = self._read_snapshots(cluster_name, cluster_cfg)
+        snap = meta.get(snapshot_name)
+        if not snap:
+            raise ConfigError(
+                f"[{cluster_name}] no snapshot '{snapshot_name}' "
+                f"(have: {', '.join(meta) or 'none'})."
+            )
+        restore_cfg = copy.deepcopy(cluster_cfg)
+        boxes = restore_cfg.get("boxes") or {}
+        for box, tag in (snap.get("boxes") or {}).items():
```
mherkazandjian, line 425: Please preflight every recorded tag with `image_exists()` before generating or bringing up the restore config. Snapshot metadata can outlive a locally pruned image; without validation Compose may try to pull the `boxman/...` tag from a registry (or fail only after starting the destructive recreate), rather than aborting with a clear invalid-snapshot error.

src/boxman/manager.py
```diff
@@ -4640,9 +4677,28 @@ def snapshot_restore(cls, cli_args):
...
+        for cname, cluster in cls._select_dc_clusters(cli_args):
+            snap = cli_args.snapshot_name
+            if not snap:
+                snaps = cls.session_for_cluster(cname).snapshot_list_cluster(cname, cluster)
+                if not snaps:
+                    cls.logger.error(f"[{cname}] no snapshots to restore")
+                    continue
+                snap = max(snaps, key=lambda k: snaps[k].get('created', ''))
+                cls.logger.info(f"[{cname}] resolved latest snapshot: '{snap}'")
+            cls.session_for_cluster(cname).snapshot_restore_cluster(cname, cluster, snap)
```
mherkazandjian: This performs the docker-compose restore before the VM snapshots below are resolved and pre-validated. In a mixed project, a missing or invalid VM snapshot then aborts after containers were already recreated, leaving a partial restore despite this method's pre-validation contract. Resolve and validate both provider types first, then mutate either one.

src/boxman/providers/docker_compose/session.py
```diff
@@ -131,6 +134,172 @@ def destroy_cluster(self, cluster_name: str, cluster_cfg: dict[str, Any]) -> boo
...
+            if not container:
+                self.logger.warning(
+                    f"[{cluster_name}] box '{box}' has no container — skipping "
+                    f"in snapshot '{snapshot_name}'."
+                )
+                continue
+            tag = _snapshot_tag(project, box, snapshot_name)
+            result = runner.commit(container, tag)
+            if not getattr(result, "ok", False):
+                raise ProvisionError(
```
mherkazandjian, line 292: [low] Partial `take` leaks untracked images: if `docker commit` fails on box N, boxes 1..N-1 are already committed, but this raise happens before `_save_snapshots()`, so those images exist on disk with no metadata entry — `snapshot delete` can never reclaim them. Consider a best-effort `image_rm` of the already-committed `tags` before raising, or recording the successful boxes in metadata so delete can clean them up.

src/boxman/providers/docker_compose/session.py
```diff
@@ -131,6 +134,172 @@ def destroy_cluster(self, cluster_name: str, cluster_cfg: dict[str, Any]) -> boo
...
+        """Remove a snapshot's images (``docker image rm``) and metadata."""
+        meta = self._read_snapshots(cluster_name, cluster_cfg)
+        snap = meta.get(snapshot_name)
+        if not snap:
+            self.logger.warning(
+                f"[{cluster_name}] no snapshot '{snapshot_name}' to delete.")
+            return False
+        runner, _wd, _cf = self._teardown_runner(cluster_name, cluster_cfg)
+        for _box, tag in (snap.get("boxes") or {}).items():
+            runner.image_rm(tag)
```
mherkazandjian: [low] The `image_rm` result is ignored and "removed snapshot image" is logged unconditionally — but `docker image rm -f` legitimately fails while a container is running from the image, e.g. `snapshot delete v1` right after `snapshot restore v1` (the restored containers hold that image). The metadata entry is still dropped, so the image leaks untracked while the log claims it was removed. Routing the result through `self._check(...)` (as the teardown ops do) would at least surface it.

src/boxman/providers/docker_compose/session.py
```diff
@@ -131,6 +134,172 @@ def destroy_cluster(self, cluster_name: str, cluster_cfg: dict[str, Any]) -> boo
...
+            if box in boxes:
+                boxes[box] = {**(boxes[box] or {}), "image": tag}
+                boxes[box].pop("build", None)  # the snapshot image replaces build
+        workdir = self._workdir(restore_cfg, cluster_name)
+        project = self._compose_project(cluster_name)
+        shared = self.config.get("shared_networks") or {}
+        compose = self._generator.generate(
+            cluster_name, restore_cfg, self._conf_dir(), shared, project_name=project)
+        compose_file = self._generator.write(compose, workdir)
+        runner = ComposeRunner(
```
mherkazandjian: [nit/reuse] Lines 258–266 re-implement `_compose_context()` with `restore_cfg` — `runner, workdir, compose_file = self._compose_context(cluster_name, restore_cfg)` looks like a drop-in (same workdir/shared-networks/project resolution, generate/write, and `ComposeRunner` construction; the restore test's `_generator`/`ComposeRunner` patches would keep passing). Reusing it also raises the parity question with `up_cluster()`, which additionally runs `_ensure_bind_dirs()` and `runner.preflight()` before `up`: as written, a bind-mount host dir deleted between take and restore gets created root-owned by the daemon during `up --force-recreate` — the exact case `_ensure_bind_dirs` exists to prevent. (`_compose_context`'s "bring-up only" docstring would want a touch-up, since restore is a bring-up.)

src/boxman/providers/docker_compose/compose_runner.py
```diff
@@ -91,6 +95,29 @@ def ps(self):
...
+        return run(
+            f"{self._sudo}docker commit {shlex.quote(container)} {shlex.quote(tag)}",
+            warn=True, hide=True)
+
+    def image_rm(self, tag: str):
+        """``docker image rm -f <tag>``."""
+        return run(f"{self._sudo}docker image rm -f {shlex.quote(tag)}",
+                   warn=True, hide=True)
+
+    def image_exists(self, tag: str) -> bool:
```
mherkazandjian, line 150: [nit] `image_exists()` has no production caller (only the test `_FakeRunner` mirrors it) — dead code as merged. It looks intended for the missing pre-check in `snapshot_restore_cluster()`: if a snapshot image was removed outside boxman, restore currently fails mid-`up --force-recreate` with docker's raw pull error. Checking each snapshot tag up front would turn that into a clean per-box error — either wire it in there or drop the helper.

src/boxman/manager.py
```diff
@@ -4573,17 +4597,30 @@ def _select_vm_targets(cls, cli_args):
...
+        docker-compose clusters are snapshotted per-cluster via ``docker
+        commit`` (decision D3; named volumes are NOT captured).
 
         Honours ``--cluster`` / ``--vms`` so a single cluster (or VM) can be
         snapshotted independently in a multi-cluster project.
         """
+        # docker-compose clusters first (cluster-scoped, D3).
+        dc_done = False
+        for cname, cluster in cls._select_dc_clusters(cli_args):
+            cls.session_for_cluster(cname).snapshot_take_cluster(
```
mherkazandjian: [low] The dc loop runs first and lets exceptions propagate, so one failing dc cluster aborts the whole command before the VM clusters (or later dc clusters) are processed. Concretely: on a mixed project whose dc cluster is deprovisioned (or defined but never up'd), `boxman snapshot take --name pre` exits 2 with "no containers to snapshot" and the VMs are never snapshotted. Likewise in `snapshot_restore` below, a `--name` that only exists on the libvirt cluster raises `ConfigError` from the dc cluster and skips the VM restore — asymmetric with the graceful `continue` used there for the unnamed/no-snapshots case, and with `snapshot_delete_cluster`'s warn+`False`. Consider catching per-cluster errors, logging, and continuing (with a non-zero exit at the end if anything failed).

src/boxman/manager.py
```diff
@@ -4514,6 +4527,17 @@ def snapshot_log(cls, cli_args):
                 f"[{vm_list}]{cur_marker}"
             )
 
+    def _select_dc_clusters(self, cli_args) -> list[tuple[str, dict]]:
```
mherkazandjian: [question] Documented-but-surprising `--vms` semantics: because the filter is ignored rather than treated as excluding containers, an explicitly narrowed `boxman snapshot restore --vms node01 --name X` on a mixed project also force-recreates every container of every dc cluster — discarding their in-container state since the snapshot — which the user arguably scoped away. Worth considering skipping dc clusters when a non-default `--vms` is passed (the default is `'all'`), or at least calling this out in the D3 doc section.

src/boxman/providers/docker_compose/session.py
```diff
@@ -367,6 +536,19 @@ def snapshot_list(self, vm_name: str | None = None) -> list[dict[str, str]]:
         self._cluster_scoped("snapshot_list")
 
 
+def _snapshot_tag(project: str, box: str, snapshot_name: str) -> str:
```
mherkazandjian, line 797: [nit] Two corners the sanitizer doesn't cover: (1) docker tags are capped at 128 chars, so an over-long snapshot name only fails later at `docker commit` with docker's raw "invalid reference format"; (2) distinct names can collide after sanitization — `v:1` and `v-1` both map to tag `v-1`, so the second `take` silently overwrites the first snapshot's image while both metadata entries survive (and deleting either removes the shared tag). A length cap plus a collision check against existing metadata (or a short hash suffix) would pre-empt both. Fine as a follow-up.

doc/docker-compose-provider/config-schema.md
```diff
@@ -241,6 +241,34 @@ destroy` removes **named** volumes (`down --volumes`) and the generated compose
 file. Bind-mount host directories are **never** removed — they are your paths
 (`./initdb`, `.`), so cleaning them up is your call.
 
+## Snapshots and container state (decision D3)
+
+`boxman snapshot` works for docker-compose clusters, backed by `docker commit`
+— with one loud caveat.
+
+- **`boxman snapshot take <name>`** commits each container in the cluster to an
```
mherkazandjian: [docs] This section shows positional names (`boxman snapshot take <name>`, `restore <name>`, `delete <name>`), but the implemented CLI takes flags: `--name <name>` (plus `-m/--description` for take) — copy-pasting these lines errors with "unrecognized arguments". Worth matching the real invocation, as the PR description does (`take --name <n>`).

mherkazandjian (commented):
**Phase 7 review — clean, well-scoped implementation of D3; everything below is minor.** Reviewed session.py / compose_runner.py, the manager wiring, the CLI parser, the tests, and the docs at full depth, including the interaction with app.py's dispatch.

**Verified correct while reviewing (no action needed):**
- Mixed-project session resolution is sound: the snapshot verbs use `session_for_cluster()`, and app.py's generic dispatch registers a session per provider type before snapshot ops run — no unregistered-session path on mixed libvirt+dc projects.
- `_snapshot_tag` matches docker's repo/tag grammar (lowercased repo charset, tag charset, leading-separator strip, non-empty fallbacks); the default timestamp snapshot name (which contains colons) sanitizes cleanly.
- `snapshot_restore_cluster` deep-copies the cluster cfg, overrides `image:`, pops `build:`, regenerates, and runs `up --force-recreate`; volume data is intentionally untouched, warned on every take, and loudly documented.
- `_read_snapshots`/`_save_snapshots` are robust to missing/corrupt metadata; `_container_names` handles both NDJSON and array-form `ps --format json` output; sudo prefixing is consistent between the new image ops and `_base()`; every new raise path surfaces as a clean exit 2 via `main()`'s `BoxmanError` handler (no tracebacks).
- The tests assert real behavior (command strings, metadata round-trip, restore cfg override, manager dc-branch routing).

**Two notes that don't need a line:**
1. `restore` leaving `<workdir>/docker-compose.yml` pinned to snapshot tags until the next `boxman up` is documented and looks acceptable. If it ever bites (a manual `docker compose up` in that window resurrects snapshot images), writing the restore compose to a side file — or regenerating the declared file after a successful restore — would close it. No change requested.
2. Rebase overlap with #70: both PRs add `ComposeRunner.ps_json` (identical command string, slightly different docstring) plus the `_FakeRunner.ps_json`/`ps_json_result` scaffolding, so the rebase conflicts there are trivial to resolve. You already flagged the rebase — just noting the exact collision points.

8 inline comments follow: 3 low-severity robustness items, the rest nits/questions/docs.

PR #73, opened 2026-07-28
feat(docker-compose): Phase 8 hardening — e2e tests, example coverage, user docs (#56)

Orski174:
Phase 8 of the docker-compose provider epic (#42) — closes #56. Branches off and targets `feat/docker-compose-provider`.

The last substantive phase: e2e coverage, an example that exercises the features Phases 5–7 added, and the user-facing documentation the provider has been missing.

## Example box

`boxes/docker-compose-standalone` becomes the **feature showcase** — a named volume (redis `/data`) and a read-only bind mount (`./site` → nginx docroot), plus README walkthroughs for volumes, `boxman exec`, the ansible inventory and snapshots. It needs no KVM and no cloud images, so the whole walkthrough runs on any docker host.

`boxes/hybrid-libvirt-docker-compose` is deliberately left alone: it teaches exactly one delicate thing (VM↔container L2 over a shared bridge), and loading it with volumes and snapshots would dilute that. Discussed with @Orski174 before implementing.

## e2e tests

New `tests/test_docker_compose_provider_e2e.py` (`make test-dc-e2e`), two tiers, both `@pytest.mark.integration` and local-only:

- **docker-only** — provision, `ps`, `exec`, published port + bind mount, `:ro` enforcement, named volume across `down`/`up`, snapshot take/restore including the volume-divergence caveat, duplicate-name rejection, and `destroy` removing containers *and* named volumes.
- **hybrid, gated on `/dev/kvm`** — mixed `ps`, an inventory spanning both providers, cluster-internal vs macvlan isolation, and VM↔container ping/ARP.

Named so it does not collide with `test_docker_compose.py`, which covers the docker-compose **runtime** — an unrelated axis that happens to share the name.

**Result on the staging host: 13 passed, 1 skipped.** The skip is the L2 ping test, which requires the VM's shared-bridge address to be assigned first (that domain has no DHCP by design); verified manually instead, see below.

## Two bugs this phase surfaced

Both were found by actually running what the docs claim, which is the point of the phase.

**1. `boxman run` never reached containers.** The task runner runs with `cwd` set to the workdir, but a relative cluster `inventory` was resolved with `normpath` only — so a docker-compose cluster's `workdir: ./.boxman/web` produced `.boxman/web/.boxman/web/inventory`. Ansible parsed no inventory and **silently fell back to an implicit localhost** — the failure mode that looks like success. Phase 6 checked the inventory's *contents* but never ran `boxman run` against it.

One-line fix in `env_loader.py` (`normpath` → `abspath`). This restores the contract the existing tests already describe as "INVENTORY (absolute)" — they passed only because their workdirs were absolute `tmp_path`s. Behaviour is identical when the workdir is already absolute. Regression test added with a relative workdir.

**2. The hybrid box README documented a command that does not exist.** `boxman ssh` takes only a VM name and `--cluster`; it opens an interactive session and accepts no trailing command. Every `boxman ssh compute_node01 -- <cmd>` line in that README failed with `unrecognized arguments`. Since "fresh-host walkthrough of the hybrid example passes" is this phase's acceptance criterion, running it is what caught this. Replaced with ssh over the `ssh_config` boxman generates in the workspace.

## Docs

- **README** — a `Providers` section: the per-cluster model, a mixed example, and an explicit disambiguation of docker-compose the *runtime* from docker-compose the *provider*, plus the two differences that bite first (`ssh` is VM-only; `--vms` is libvirt-only).
- **`doc/docker-compose-provider/user-guide.md`** — the practical path: prerequisites, minimal project, lifecycle table, volumes, networking, `exec`, ansible, snapshots, mixed-project scoping, troubleshooting.
- **`doc/docker-compose-provider/migration-v1-to-v2.md`** — v1.0 → v2.0, quoting the errors boxman actually emits (checked against the code, not assumed — e.g. it is the docker-compose provider specifically that requires v2.0, not any per-cluster `provider:` key).
- **design.md** — an AC1–AC18 coverage table naming the test that proves each one; the doc index gains a user-facing section and a phase status table, replacing a stale "Begin Phase 1" next-steps list.

One limitation is now documented rather than papered over: ansible **modules** need a Python interpreter inside the container, which alpine-based images lack. `ansible -m raw` works (verified); `boxman run --cmd` wraps `ansible.builtin.shell` and so needs an image with Python. An earlier draft of the box README claimed `boxman run --cmd 'nginx -v'` just works — it does not, and that claim was corrected.

## Verification

- **Host unit suite: 1329 green** (+1 for the env_loader regression).
- **e2e on the staging host: 13 passed, 1 skipped.**
- **Live, with the L2 precondition met** (AC10/AC11/AC12): ping 0% loss in both directions; ARP resolving the VM to `52:54:00:aa:00:02` — `adapter_2`'s MAC straight out of the box conf, so this is L2 and not routing — and the container's cluster-internal address unreachable from the VM (100% loss).
- Every markdown link added here resolves; the generated compose for the enriched box was checked through the real `ComposeGenerator` (named volume + top-level declaration, bind resolved to an absolute path with `:ro`).

## Not included

- `README.md` has a pre-existing broken link, `[MIT License](../LICENSE)`, which should be `LICENSE` — left alone as unrelated to this phase.

### DIFF
--- a/src/boxman/utils/env_loader.py
+++ b/src/boxman/utils/env_loader.py
@@ -185,6 +185,11 @@ def load_workspace_env(
                 # an implicit localhost. Anchoring to the cwd at load time (the
                 # project root) makes the value immune to the later chdir; when
                 # the workdir is already absolute this is identical to before.
+                # This applies to every path override (INVENTORY, SSH_CONFIG,
+                # ANSIBLE_CONFIG) — all are consumed by those same chdir'd
+                # subprocesses. Workspace-level keys are unaffected: they come
+                # through `_apply(workspace_config, None)`, where resolve_base
+                # is falsy, so they keep their historical relative handling.
                 val = os.path.abspath(os.path.join(resolve_base, val))
             env_vars[env_key] = val
             # ansible consults ANSIBLE_INVENTORY (which env.sh seeds from
--- a/src/boxman/utils/env_loader.py
+++ b/src/boxman/utils/env_loader.py
@@ -177,7 +177,15 @@ def load_workspace_env(
             # point `boxman ssh` / `run` at a file that was never created.
             resolve_base = ssh_base if (env_key == "SSH_CONFIG" and ssh_base) else base
             if resolve_base and not os.path.isabs(val):
-                val = os.path.normpath(os.path.join(resolve_base, val))
+                # abspath, not normpath: tasks run with cwd set to the workdir
+                # (or workspace.path), so a result that is still relative would
+                # be resolved a *second* time against that cwd and miss. With a
+                # relative cluster workdir like './.boxman/web' that produced
+                # '.boxman/web/.boxman/web/inventory' and ansible fell back to
+                # an implicit localhost. Anchoring to the cwd at load time (the
+                # project root) makes the value immune to the later chdir; when
+                # the workdir is already absolute this is identical to before.
+                val = os.path.abspath(os.path.join(resolve_base, val))
             env_vars[env_key] = val
             # ansible consults ANSIBLE_INVENTORY (which env.sh seeds from
             # INVENTORY); keep it in step so a repointed inventory actually
--- a/tests/test_docker_compose_provider_e2e.py
+++ b/tests/test_docker_compose_provider_e2e.py
@@ -202,9 +202,15 @@ class TestDockerComposeProviderLifecycle:
 
         assert _boxman("snapshot restore --name e2e1", cwd=standalone_up, warn=True).ok
 
-        # the container filesystem went back to the snapshot
+        # the container filesystem went back to the snapshot. Assert the
+        # specific "no such file" rather than merely a non-zero exit, which a
+        # broken `exec` (or a container left down by the restore) would also
+        # produce — that would turn a failed restore into a green test.
         gone = _boxman("exec web.cache -- ls /tmp/dirt", cwd=standalone_up, warn=True)
         assert not gone.ok, "container filesystem was not rolled back"
+        assert "No such file" in (gone.stdout + gone.stderr), (
+            f"expected /tmp/dirt to be absent after restore, but the command "
+            f"failed for a different reason:\n{gone.stdout}{gone.stderr}")
 
         # ...but the volume kept the *post-snapshot* write (docker commit never
         # captured it), which is the divergence the docs warn about
@@ -381,8 +387,15 @@ class TestHybridVmContainer:
     def test_cluster_internal_address_unreachable_from_vm(self, hybrid_up):
         """AC12, the traffic-level half: the container's cluster-internal
         address is not reachable from the VM — only the shared bridge is
-        L2-adjacent."""
+        L2-adjacent.
+
+        Asserts ping actually ran and reported total loss. A bare
+        ``not out.ok`` would also be satisfied by ssh failing outright, which
+        would pass this test without ever demonstrating isolation.
+        """
         out = _vm_ssh(hybrid_up, "ping -c2 -W2 172.31.0.2", warn=True)
-        assert not out.ok, (
-            "the cluster-internal network is reachable from the VM — isolation "
-            f"regression:\n{out.stdout}")
+        assert "100% packet loss" in out.stdout, (
+            "expected total loss to the cluster-internal address; either it is "
+            f"reachable (isolation regression) or ping never ran:\n"
+            f"{out.stdout}{out.stderr}")
+        assert not out.ok
--- a/tests/test_docker_compose_provider_e2e.py
+++ b/tests/test_docker_compose_provider_e2e.py
@@ -19,8 +19,10 @@ Two tiers:
 These are local-only, like ``make test-provision`` — they create and destroy
 real infrastructure::
 
-    make test-dc-e2e                                  # docker-only tier
-    make test-dc-e2e pytest_args="-k hybrid"          # hybrid tier too
+    make test-dc-e2e                              # both tiers; hybrid
+                                                  # auto-skips without /dev/kvm
+    make test-dc-e2e pytest_args="-k Hybrid"      # restrict to the hybrid tier
+    make test-dc-e2e pytest_args="-k Lifecycle"   # restrict to the docker-only tier
 """
 
 import json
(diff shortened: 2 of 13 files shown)


tests/test_docker_compose_provider_e2e.py
```diff
@@ -0,0 +1,300 @@
...
+
+    def test_snapshot_take_rejects_duplicate_name(self, standalone_up):
+        assert _boxman("snapshot take --name dup", cwd=standalone_up, warn=True).ok
+        again = _boxman("snapshot take --name dup", cwd=standalone_up, warn=True)
+        assert "already exists" in (again.stdout + again.stderr)
+        _boxman("snapshot delete --name dup", cwd=standalone_up, warn=True)
+
+    def test_destroy_removes_containers_and_named_volumes(self):
+        """AC15: destroy tears down containers *and* named volumes."""
+        _boxman("provision --force", cwd=STANDALONE, warn=True)
```
mherkazandjian: **nit (test robustness):** this `provision --force` result is never asserted. If the re-provision fails (a transient docker hiccup, say), `destroy` of an absent project still succeeds and both emptiness assertions pass vacuously — AC15 goes green without ever being exercised. Suggest `assert _boxman("provision --force", cwd=STANDALONE, warn=True).ok` like the fixture does.

Related, non-blocking: this test deliberately doesn't take `standalone_up` yet destroys the module-scoped deployment, so it must stay the *last* test in definition order — anything added after it (or a reordering plugin like `pytest-randomly`) would run against a destroyed project. The double destroy with the fixture teardown is harmless (`warn=True`, idempotent), but a one-line comment stating the ordering contract would keep it from being broken accidentally.

tests/test_docker_compose_provider_e2e.py
```diff
@@ -0,0 +1,300 @@
...
+  persistence across down/up, snapshot take/restore/delete, destroy. Needs
+  Docker + compose v2 only, so it runs anywhere Docker does.
+* **hybrid (KVM-gated)** — ``boxes/hybrid-libvirt-docker-compose``: a libvirt
+  VM and a container sharing an L2 domain, plus the mixed-project ps and
+  inventory. Skipped without ``/dev/kvm``.
+
+These are local-only, like ``make test-provision`` — they create and destroy
+real infrastructure::
+
+    make test-dc-e2e                                  # docker-only tier
```
mherkazandjian: **nit (docstring accuracy):** `-k hybrid` *restricts* the run to the hybrid tier — it doesn't add it on top. Plain `make test-dc-e2e` already runs both tiers (the hybrid one auto-skips without `/dev/kvm` via `requires_kvm`), so the two annotations have it backwards: the first line runs everything available on the host, this one runs hybrid-only.

src/boxman/utils/env_loader.py
```diff
@@ -177,7 +177,15 @@ def _apply(source: dict, base: str | None, ssh_base: str | None = None) -> None:
...
-                val = os.path.normpath(os.path.join(resolve_base, val))
+                # abspath, not normpath: tasks run with cwd set to the workdir
+                # (or workspace.path), so a result that is still relative would
+                # be resolved a *second* time against that cwd and miss. With a
+                # relative cluster workdir like './.boxman/web' that produced
+                # '.boxman/web/.boxman/web/inventory' and ansible fell back to
+                # an implicit localhost. Anchoring to the cwd at load time (the
+                # project root) makes the value immune to the later chdir; when
+                # the workdir is already absolute this is identical to before.
+                val = os.path.abspath(os.path.join(resolve_base, val))
```
mherkazandjian, line 193: **NOTE (verified — no action needed):** confirmed correct against the task-runner path: subprocesses run with `cwd=workdir` and inherit `INVENTORY`, so the relative result really was resolved twice (`.boxman/web/.boxman/web/inventory`), and `abspath(join(base, val))` is byte-identical to the old `normpath(...)` whenever `base` is absolute — no behaviour change for existing absolute-workdir projects.

One observation the comment doesn't spell out: this branch also covers `SSH_CONFIG` and `ANSIBLE_CONFIG` (all of `path_overrides`), not just the inventory the comment narrates. That is equally right — they're consumed by the same chdir'd subprocesses — and workspace-level keys keep their historical handling since `_apply(workspace_config, None)` leaves `resolve_base` falsy. The regression test reproduces the original failure cleanly.

mherkazandjian (commented):
## Review — Phase 8 hardening

Reviewed at full rigor against `feat/docker-compose-provider` (base), checking the docs' code-level claims against the actual code rather than the prose. **Verdict: no blocking issues.** The single production change is correct, the docs are unusually accurate, and the findings below are nits/notes only.

### The `env_loader.py` fix — verified correct

Tasks execute subprocesses with `cwd=workdir` and `INVENTORY` from `load_workspace_env` flows into them, so a still-relative value got resolved a *second* time against the per-task cwd — the `.boxman/web/.boxman/web/inventory` double path, after which ansible silently fell back to implicit localhost. `abspath` anchors at load time (cwd = project root, before any per-task chdir), and `abspath(join(base, val))` is byte-identical to the old `normpath(...)` whenever the base is absolute — no behaviour change for existing projects. The regression test genuinely reproduces the bug (`monkeypatch.chdir` + relative `./.boxman/web` workdir). Scope detail in the inline note.

### Doc claims spot-checked against base code — all accurate

- **CLI**: `ssh` takes only a positional VM name + `--cluster` — no trailing command (`cli_parser.py`), so the hybrid README correction is right. `exec` has `--shell` and `-- cmd`; `ps --json`, `provision --force`, `destroy --auto-accept`, `snapshot take --name/-m` all exist as used.
- **`--vms` is libvirt-only**: a narrowed `--vms` deselects dc clusters with an explicit warning (`manager.py:300–314`) — README/user-guide/migration wording matches the code.
- **Quoted error strings match verbatim**: `unsupported config version: '…' (supported: '1.0', '2.0')`; `cluster '…' uses the docker-compose provider, which requires version: '2.0'`; `declares both 'boxes:' and 'vms:' — use one (prefer 'boxes:' in v2.0).`; `snapshot '…' already exists`; `no containers to snapshot`; `'docker compose' (Compose v2 plugin) is not available`; the `requires runtime 'local'` guardrail. `version: 2`/`'2.0'` both select v2.0, and the legacy `vms:` deprecation warning exists.
- **Lifecycle table ≡ `session.py`**: down→`stop`, deprovision→`down` (named volumes kept), destroy→`down --volumes`, suspend/resume→`pause`/`unpause`, `control save` explains-and-skips (`manager.py:5347`); `up -d --wait`, `readiness_timeout` default 120.
- **Snapshots**: image repo `boxman/<project>_<cluster>_<box>:<name>` ≡ `compose_project_name()` (`<project_name>_<cluster>`) + box; `snapshots.json` in the cluster workdir; take-time volume warning; the volume-divergence caveat matches `docker commit` semantics.
- **Volumes**: the example's `{name, container_path}` / `{host_path, container_path, readonly}` match `VALID_VOLUME_KEYS`; `size:` advisory warning, box `workdir` = bind of `.`, host-dir pre-created as the invoking user — all present in `compose_generator.py`.
- **Networking**: macvlan parented on the shared host bridge; `disable_netfilter: true` exists and is loudly discouraged (`netlab/shared_bridges.py:146`).
- **Example-box facts**: compose project `dc_standalone_web`, container `hybrid_libvirt_dc_services-web-1`, VM MAC `52:54:00:aa:00:02` = `adapter_2`, bridge `bx_app`, `WS=~/workspaces/boxmandev/hybrid-libvirt-docker-compose` — all match the confs.
- **design.md AC table**: every referenced unit/e2e test name exists (base or this PR).
- **Links**: every added markdown link resolves, including the `#first-the-name-collision` / `#snapshots` anchors; `config-schema.md`, `adr-001-per-cluster-provider.md`, `spike/findings.md` all exist.

### Findings (ranked — none blocking)

1. **nit** — the e2e destroy test never asserts its own re-provision, so it can pass vacuously (inline).
2. **nit** — the e2e module docstring's `-k hybrid` annotation is inverted (inline).
3. **NOTE** — env_loader scope observation, verified fine, no action (inline).

The "run what the docs claim" discipline clearly paid off here — nice work.

tests/test_docker_compose_provider_e2e.py
```diff
@@ -0,0 +1,300 @@
...
+        # dirty the container filesystem after the snapshot
+        _boxman("exec web.cache -- sh -c 'echo dirt > /tmp/dirt'", cwd=standalone_up)
+        assert _boxman("snapshot restore --name e2e1", cwd=standalone_up, warn=True).ok
+
+        gone = _boxman("exec web.cache -- ls /tmp/dirt", cwd=standalone_up, warn=True)
+        assert not gone.ok, "container filesystem was not rolled back"
+
+        # ...while the named volume is untouched by the snapshot
+        kept = _boxman("exec web.cache -- redis-cli get persisted", cwd=standalone_up)
+        assert "yes" in kept.stdout
```
mherkazandjian: **test correctness (worth fixing):** this doesn't actually prove the documented volume divergence. The key is written *before* `snapshot take` and never changed afterward, so `yes` is the expected result whether the named volume was correctly left untouched **or** incorrectly captured/restored with the container — the assertion can't fail for the reason the test exists. Mutate the key *after* the snapshot (e.g. `before` → `after`) and assert the post-snapshot value survives the restore; that version actually fails if volume data were rolled back.

tests/test_docker_compose_provider_e2e.py
```diff
@@ -0,0 +1,300 @@
...
+        """AC10/AC11: the macvlan container pings the VM on the shared bridge
+        and learns its MAC via ARP.
+
+        The VM's shared-bridge address is assigned by the README walkthrough
+        (that L2 domain has no DHCP), so skip rather than fail when it has not
+        been set up — this test asserts boxman's plumbing, not the operator's.
+        """
+        ping = _boxman(
+            "exec services.web -- ping -c2 -W2 10.10.0.20", cwd=hybrid_up, warn=True)
+        if not ping.ok:
```
mherkazandjian: **test correctness:** skipping on any failed ping makes this unable to detect the AC10/AC11 regressions it is meant to cover — a broken macvlan parent, missing bridge attachment, firewall regression, or a wrong container address all produce the same `ping.ok == False` path and turn the test green. The fixture should configure `10.10.0.20` on the VM first (or explicitly assert that only the address precondition is absent), then treat a ping failure as a test failure rather than a skip.

mherkazandjian (commented):
Consolidating two findings that surfaced during this review but landed under empty-body review shells (a review-tooling glitch) — re-posting them here so they're attached to the main review [#pullrequestreview-4799334177]. Both are test-quality issues on the new e2e suite. Neither blocks merge, but the first is worth fixing: as written, that test can't fail for the reason it exists.

PR #77, opened 2026-08-02
feat(net): support static dhcp reservations in libvirt networks

mherkazandjian:
## what

A libvirt network could only declare a dynamic DHCP range, so a guest's address
was whatever dnsmasq happened to hand out, and it moved across rebuilds. There
was no way to pin one.

This adds an optional `ip.dhcp.hosts` list:

```yaml
networks:
  mgmt:
    ip:
      address: '10.5.3.1'
      netmask: '255.255.255.0'
      dhcp:
        range:
          start: '10.5.3.50'
          end: '10.5.3.100'
        hosts:
          - mac: '52:54:00:0c:01:01'
            name: ctrl-1-01
            ip: '10.5.3.10'
```

which renders to

```xml
<dhcp>
  <range start='10.5.3.50' end='10.5.3.100'/>
  <host mac='52:54:00:0c:01:01' name='ctrl-1-01' ip='10.5.3.10'/>
</dhcp>
```

The optional `name` becomes the dnsmasq hostname, so the guest also resolves by
name for everything else on that network.

## why the validation

libvirt rejects a malformed reservation with an error that does not say *which*
entry is at fault, and some mistakes are accepted and then silently do the wrong
thing. `_parse_dhcp_hosts` fails fast on:

- an entry missing `mac` or `ip`
- an `ip` that does not parse
- an `ip` outside the network the reservation belongs to
- the same mac, or the same ip, reserved twice

Macs are compared case-folded — otherwise the same address written in mixed case
slips through as two entries and the last one silently wins.

## notes

- A reservations-only network — every guest pinned, no dynamic pool — is valid
  libvirt and now renders correctly instead of emitting an empty `<ip>`.
- The `<host>` loop is emitted after `<range>` because libvirt's schema requires
  that order.
- A reservation may sit inside or outside the dynamic range; dnsmasq excludes it
  from the pool either way. Keeping them disjoint is just clearer.
- The mac is necessarily repeated between the reservation and the VM's adapter.
  libvirt keys on the mac and boxman does not derive one from the other, so a
  stale entry fails silently by falling back to the pool. Deriving reservations
  from adapters automatically would remove the duplication but adds magic, so
  it is left explicit.

## tests

12 new tests in `tests/test_libvirt_net.py::TestDhcpReservations` covering
rendering, the optional name, reservations without a range, mac normalisation,
element ordering, and each validation failure.

Full suite: 1147 passed, 0 failures.

### DIFF
--- a/boxes/multi-cluster-routed-via-frr/conf.yml
+++ b/boxes/multi-cluster-routed-via-frr/conf.yml
@@ -181,6 +181,15 @@ clusters:
             range:
               start: '192.168.10.2'
               end: '192.168.10.254'
+            # pin a guest to a fixed address by its mac. the mac has to match
+            # the vm's network_adapters entry below, and the optional name
+            # becomes its dnsmasq hostname on this network. a reservation may
+            # sit inside the dynamic range -- dnsmasq excludes it from the pool
+            # either way -- but narrowing the range to leave room is clearer
+            #hosts:
+            #  - mac: '52:54:00:11:01:01'
+            #    name: app1
+            #    ip: '192.168.10.10'
         enable: True
         autostart: True
 
--- a/src/boxman/assets/network.xml.j2
+++ b/src/boxman/assets/network.xml.j2
@@ -11,9 +11,12 @@
     {% if dhcp_range_start and dhcp_range_end %}
       <range start='{{ dhcp_range_start }}' end='{{ dhcp_range_end }}'/>
     {% endif %}
-    {# libvirt's schema wants <range> before <host>, so this loop stays last #}
+    {# <range> first, then <host>: libvirt accepts either order but emits and
+       documents this one, so a dumpxml diff stays readable. the attributes are
+       escaped because the environment has no autoescape and a reservation name
+       is free-form config text #}
     {% for host in dhcp_hosts %}
-      <host mac='{{ host.mac }}'{% if 'name' in host %} name='{{ host.name }}'{% endif %} ip='{{ host.ip }}'/>
+      <host mac='{{ host.mac|e }}'{% if 'name' in host %} name='{{ host.name|e }}'{% endif %} ip='{{ host.ip|e }}'/>
     {% endfor %}
     </dhcp>
   {% endif %}
--- a/src/boxman/assets/network.xml.j2
+++ b/src/boxman/assets/network.xml.j2
@@ -6,9 +6,15 @@
   <bridge name='{{ bridge_name }}' stp='{{ bridge_stp }}' delay='{{ bridge_delay }}'/>
   <mac address='{{ mac_address }}'/>
   <ip address='{{ ip_address }}' netmask='{{ netmask }}'>
-  {%- if dhcp_range_start and dhcp_range_end %}
+  {% if (dhcp_range_start and dhcp_range_end) or dhcp_hosts %}
     <dhcp>
+    {% if dhcp_range_start and dhcp_range_end %}
       <range start='{{ dhcp_range_start }}' end='{{ dhcp_range_end }}'/>
+    {% endif %}
+    {# libvirt's schema wants <range> before <host>, so this loop stays last #}
+    {% for host in dhcp_hosts %}
+      <host mac='{{ host.mac }}'{% if 'name' in host %} name='{{ host.name }}'{% endif %} ip='{{ host.ip }}'/>
+    {% endfor %}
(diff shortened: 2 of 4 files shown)


mherkazandjian (commented):
## Review — approve with changes

Well-scoped and genuinely well-tested. I verified the mechanism against libvirt 11.2.0 rather than reading only: each case below was rendered through this PR's own template and run through `virsh net-define` (defined, then undefined; never started, so no bridges were created).

### Verified working

- Valid reservation, optional `name`, and the reservations-only-no-range case all define cleanly. The empty-`<ip>` problem described in the body is real and correctly fixed.
- Rendered XML is well-formed with no stray blank lines — `generate_xml` sets `trim_blocks`/`lstrip_blocks`, so dropping the `{%-` was safe.
- `import ipaddress` is already on main, and `self.logger` is set by `super().__init__()` before `_parse_dhcp_hosts()` runs — no ordering hazard.
- Suite reproduces: **1147 passed, 108 deselected**.

### 1. Gateway / network / broadcast addresses pass validation — medium

`_parse_dhcp_hosts` checks `address not in network`, which accepts `10.5.3.1` (the bridge's own address), `10.5.3.0` and `10.5.3.255`. libvirt **accepts all three at define time** (verified). Handing a guest the gateway's address is exactly the "defines cleanly and then silently does the wrong thing" class this PR sets out to catch — and it is the most likely mistake, since the gateway is the address already sitting at the top of the same YAML block.

```python
interface = ipaddress.IPv4Interface(f"{self.ip_address}/{self.netmask}")
network = interface.network
...
if address == interface.ip:
    raise ValueError(f"network {self.name}: the dhcp reservation {ip} for {mac} "
                     f"is the network's own gateway address")
if address in (network.network_address, network.broadcast_address):
    raise ValueError(f"network {self.name}: the dhcp reservation {ip} for {mac} "
                     f"is the network/broadcast address of {network}")
```

### 2. Duplicate `name` is not checked — low/medium

`mac` and `ip` are deduped; `name` is the third identity field and is not. Two reservations sharing a hostname define fine (verified), and the body sells `name` as "the guest also resolves by name" — so a copy-paste that duplicates the hostname degrades name resolution silently. Same `seen_*` treatment, case-folded like the mac.

### 3. `hosts:` written as a mapping crashes unhelpfully — low

```yaml
hosts:
  mac: '52:54:00:0c:01:01'   # forgot the '-'
  ip: '10.5.3.10'
```

gives `AttributeError: 'str' object has no attribute 'get'` — no network name, no hint. Every other failure in this function produces a precise `ValueError`. An `isinstance(hosts_info, list)` guard (and `isinstance(entry, dict)`) closes it.

### 4. The `<range>`-before-`<host>` justification is not accurate — low, comments only

The template comment says *"libvirt's schema wants `<range>` before `<host>`"* and the test comment says *"a reversed document is rejected at define time"*. On libvirt 11.2.0 a reversed document **passes `virt-xml-validate` and defines successfully** — both tested. Emitting `<range>` first is still right (it matches libvirt's own output), but the stated reason is not true, and `test_ordering_puts_range_before_hosts` asserts a house style rather than a libvirt constraint. Worth rewording both so nobody later "fixes" a constraint that does not exist.

### 5. MAC format is not validated — nit, arguably fine as-is

`ip` is parsed, `mac` is not. But libvirt's own error here is clear and quotes the offender: `XML error: Cannot parse MAC address 'not-a-mac' in network 'demo'`. This field is a genuine exception to the PR's rationale — probably a sentence in the body rather than code.

### 6. `seen_ips` keys on the raw string — nit

`address` is already a canonical `IPv4Address`; keying the dedup map on `str(address)` instead of the unparsed `ip` is free and removes any dependence on input spelling.

### Follow-up (not a blocker)

`define_networks()` is called only from the provision path (`manager.py:3230`); `update()` does not reconcile networks. Adding a reservation to an already-provisioned project therefore does nothing until a deprovision/provision or `up --force`. Consistent with all other network config, but reservations are the kind of knob users will expect `boxman update` to pick up — worth a line in the body or a follow-up issue.

Also: no `boxes/` example uses `hosts:` yet. The body's YAML would drop nicely into one of the multi-cluster boxes.

### Style

Matches the codebase well — lowercase log/error text, `Args/Returns/Raises` docstring, walrus for the optional name, errors that name both colliding entries. The mac case-folding and its comment are a good catch. Tests are readable and the parametrized failure table is the right shape.

PR #143, opened 2026-08-13
feat(libvirt): virtio-balloon free-page-reporting (memballoon)

mherkazandjian:
Closes #142.

## What

Per-VM `memballoon:` config that enables virtio-balloon **free-page reporting** so guests continuously return freed pages to the host (plus optional balloon stats period):

```yaml
vms:
  node01:
    memory: 2048
    memballoon:
      free_page_reporting: true
      stats_period: 5
```

## How

- `VirshEdit.apply_memballoon_to_xml()` — pure lxml helper (ensures `<memballoon model='virtio'>`, sets the `freePageReporting` attribute + `<stats period='N'/>`)
- `VirshEdit.configure_memballoon()` / `LibVirtSession.configure_vm_memballoon()` — persistent-config edit + redefine
- wired into provision (`_configure_and_start_vm`) and `boxman update` (idempotent; no differ changes)
- VirtualBox: Phase-2 stub; direct (PXE/ISO) VMs covered via the same configure path
- `scripts/membench.sh` — A/B measurement helper (QEMU RSS around an in-guest allocate+free churn)
- docs: new "Returning unused memory to the host" section in the tutorial

## Verification

- 8 new unit tests in `tests/test_libvirt_virsh_edit.py`; full default pytest tier green (1900 passed)
- End-to-end on hp3 (6× 1 GiB Ubuntu 24.04 clones, 600 MiB in-guest churn):
  - without memballoon: RSS stays elevated, **0% reclaimed**
  - with `free_page_reporting: true`: **94–98% reclaimed** within 90s

Caveat baked into docs: changes to a running VM land in the persistent config and take effect at next boot (QEMU device property). `freePageReporting` is an attribute of `<memballoon>` — the child-element form is silently dropped by libvirt.


src/boxman/manager_parts/vms.py
```diff
@@ -555,7 +567,9 @@ def _update_single_vm(
                 diff['changed_shared_folders']
             )
 
-            if not has_changes:
+            memballoon_config = vm_info.get('memballoon')
+
+            if not has_changes and not memballoon_config:
```
mherkazandjian: **P2 — Diff the balloon state instead of treating configuration presence as a change.** With any truthy `memballoon` block, this condition is false on every run, so `boxman update --dry-run` always reports `memballoon: apply config`; a normal update always calls `virsh define` and marks the VM updated even when the inactive XML already matches. The opposite transition is also missed: removing the block after enabling it takes this early return and leaves `freePageReporting`/`stats` behind. Please reconcile the actual inactive balloon XML against the desired/default state and only report/apply when it differs.

doc/tutorial/README.md
```diff
@@ -531,6 +531,31 @@ exit
...
+
+```yaml
+      node01:
+        memory: 2048
+        memballoon:
+          free_page_reporting: true   # guest returns freed pages to the host
+          stats_period: 5             # optional; balloon stats sample interval (s)
+```
+
+No guest-side setup is needed beyond a Linux kernel >= 5.7 (all supported cloud
```
mherkazandjian, line 549: **P2 — Document/check the host-side minimum versions.** The tutorial currently names only Linux >= 5.7, but the [libvirt domain XML reference](https://libvirt.org/formatdomain.html#memory-balloon-device) marks `freePageReporting` as available since libvirt 6.9.0, and #142 itself requires QEMU >= 5.1. On older distro packages `virsh define` can reject or drop this while `_configure_and_start_vm` logs a warning and continues booting, leaving the user without reclaim. Please list the libvirt/QEMU minimums and ideally show how to verify the resulting inactive XML.

src/boxman/providers/libvirt/virsh_edit.py
```diff
@@ -244,6 +244,82 @@ def _drop_topology(xml_content: str) -> str:
...
+                memballoon.set('model', 'virtio')
+        else:
+            devices = tree.xpath('//devices')
+            if not devices:
+                raise ProvisionError("domain xml has no <devices> element")
+            memballoon = etree.SubElement(devices[0], 'memballoon')
+            memballoon.set('model', 'virtio')
+
+        if 'free_page_reporting' in config:
+            memballoon.set('freePageReporting',
```
mherkazandjian, line 316: **P3 — Validate config values before converting them.** There is no schema/type check before here: YAML `free_page_reporting: "false"` is a truthy string and enables reporting, while `stats_period: true` becomes `1` and a float is silently truncated by `int()`. Please require an actual bool and a non-negative integer (excluding bool for the period) and surface a clear config error, so malformed input cannot silently produce the opposite setting.

scripts/membench.sh
```diff
@@ -0,0 +1,221 @@
...
+    # QEMU process RSS in MiB. Prefer virsh dommemstat (host-side counter),
+    # fall back to ps on the qemu process.
+    local dom="$1" rss
+    rss=$($VIRSH_CMD dommemstat "$dom" 2>/dev/null | awk '/^rss / {print int($2/1024)}' || true)
+    if [[ -z $rss ]]; then
+        rss=$(ps -o rss= -C qemu-system-x86 -C qemu-system-x86_64 -C qemu-kvm \
+              --no-headers 2>/dev/null | head -1 | awk '{print int($1/1024)}' || true)
+    fi
+    echo "${rss:-NaN}"
+}
```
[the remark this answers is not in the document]
mherkazandjian: Fixed in f72517f — the fallback now resolves the domain's own QEMU process via `pgrep -f "guest=${dom},"` before reading RSS, so other VMs' processes can't leak into a per-VM measurement.

scripts/membench.sh
```diff
@@ -0,0 +1,221 @@
+#!/usr/bin/env bash
+# membench.sh — measure how much host RAM boxman VMs actually hold onto
+# after a guest allocates and frees memory (virtio-balloon free-page-reporting
+# A/B test helper).
```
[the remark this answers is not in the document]
mherkazandjian: Fixed in f72517f — header rewritten as a single balanced sentence.

src/boxman/manager_parts/vms.py
mherkazandjian: Addressed in f72517f. `VMStateDiffer` now reads the actual inactive balloon state (`get_actual_memballoon`) and compares it against the normalized desired state (`normalize_memballoon_config`, absent block = libvirt defaults). `boxman update` on a matching state reports "no changes", and removing the block reconciles to `freePageReporting='off'` + stats removed. Verified live on hp3: enable → no-op → remove → reconcile → no-op.

doc/tutorial/README.md
mherkazandjian: Done in f72517f — the tutorial now lists libvirt >= 6.9.0 and QEMU >= 5.1 as host minimums and shows a `virsh dumpxml --inactive <vm> | grep -A2 memballoon` verification snippet.

src/boxman/providers/libvirt/virsh_edit.py
mherkazandjian: Done in f72517f — `validate_memballoon_config` requires a real bool for `free_page_reporting` and a positive int (bool/float/string rejected) for `stats_period`, raising `ConfigError`; `configure_memballoon` re-raises it instead of converting to False, so malformed conf.yml fails loudly. Covered by new unit tests.

PR #158, opened 2026-08-18
fix(libvirt): routed-network isolation that reconciles and allows DHCP, plus destroy round-trip fixes

### LINKED ISSUE #152: route networks: complete-isolation rules block DHCP, so dhcp: ranges on mode: route can never work
## Summary

`mode: route` networks install "complete isolation" firewall rules that also block DHCP, so a guest
NIC on a routed network can never obtain a lease. Every `dhcp:` block on a `mode: route` network in
`boxes/` is therefore dead config: those NICs come up link-local only.

Found while getting `boxes/tiny-libvirt-arch-cloudinit` working (#148). Anchors against `dcde8a1`.

## Mechanism

`Network.apply_route_iptables_rule` (`src/boxman/providers/libvirt/net.py`) installs three rules per
routed network, documented as "allow vm-to-vm communication on the same bridge / block all traffic
between host and guests in both directions":

```
-I FORWARD -i virbrN -o virbrN -j ACCEPT
-I INPUT   -i virbrN -j DROP
-I OUTPUT  -o virbrN -j DROP
```

DHCP is not *forwarded* traffic — it terminates **on the host**. libvirt runs a dnsmasq bound to the
bridge address (e.g. `10.0.14.1`), so:

* the guest's `DHCPDISCOVER` arriving on `virbrN` is addressed to the host → **INPUT** → DROP
* dnsmasq's `DHCPOFFER` leaving the host → **OUTPUT** → DROP

The `FORWARD ... ACCEPT` rule does not help: it only covers packets crossing the bridge *between
guests*, never packets to or from the host itself. And because the rules are inserted with `-I`, they
sit at the top of their chains, ahead of anything libvirt might add.

## Demonstrated

Same guest, same NIC, same command; the only variable is the two DROP rules.

**With the rules in place** — 5 discovers, zero offers, timeout:

```
DHCPDISCOVER on eth1 to 255.255.255.255 port 67 interval 4
DHCPDISCOVER on eth1 to 255.255.255.255 port 67 interval 4
DHCPDISCOVER on eth1 to 255.255.255.255 port 67 interval 8
DHCPDISCOVER on eth1 to 255.255.255.255 port 67 interval 8
DHCPDISCOVER on eth1 to 255.255.255.255 port 67 interval 10
exit=124
```

The INPUT DROP counter went `3 → 8` over that run — exactly +5, one per discover. The packets were
dropped at INPUT, not lost elsewhere.

**With the two DROP rules temporarily removed** — immediate success:

```
DHCPDISCOVER on eth1 to 255.255.255.255 port 67 interval 5
DHCPOFFER of 10.0.14.152 from 10.0.14.1
DHCPREQUEST for 10.0.14.152 on eth1 to 255.255.255.255 port 67
DHCPACK of 10.0.14.152 from 10.0.14.1
bound to 10.0.14.152 -- renewal in 1406 seconds.
exit=0
```

and the lease appeared in `virsh net-dhcp-leases`. dnsmasq was running and willing the whole time —
it answered instantly once the rules came off. (Rules restored afterwards.)

## Impact

Any box with a `mode: route` network that declares a `dhcp:` range — which is most of them, since the
pattern was copied between boxes:

* `tiny-libvirt-ubuntu-24.04-cloudinit`, `tiny-libvirt-rocky-9-cloudinit`,
  `tiny-libvirt-centos-7-cloudinit`, `tiny-multi-cluster-ubuntu-24.04-cloudinit`,
  `ten-tiny-libvirt-ubuntu-24.04-cloudinit`, and the new `tiny-libvirt-arch-cloudinit`.

It has gone unnoticed because nothing verifies the second NIC — `_verify_ssh_connection` only ever
uses the first. The guest boots, SSH works over the NAT NIC, and the isolated NIC quietly sits at
`degraded (configuring)` with a link-local address.

## Suggested fix

The isolation intent is "guests cannot talk to the host". DHCP can be excepted without giving that up,
because a DHCP lease does not let the guest reach anything else on the host. Punch a DHCP-shaped hole
above the DROPs — the same thing libvirt does for its own networks:

```
-I INPUT  -i virbrN -p udp --dport 67 -j ACCEPT   # guest -> dnsmasq
-I OUTPUT -o virbrN -p udp --dport 68 -j ACCEPT   # dnsmasq -> guest
```

Note ordering: `_ensure_rule` uses `-I`, so whichever rule is inserted last ends up on top. The ACCEPT
pair must be inserted *after* the DROP pair to sit above it.

Whether to extend the same exception to DNS (dnsmasq also serves `53` on that address) is a policy
call — DHCP-only keeps the isolation tightest and is enough to make the declared `dhcp:` ranges work.

Alternatives, if the current behaviour is considered correct:

* reject `dhcp:` on a `mode: route` network at config-validation time, with a message pointing at a
  static address in `cloudinit_network_config`, and
* fix the existing boxes, which currently advertise a DHCP range that cannot work.

Either way the present state — accepting the config, starting a dnsmasq for it, and silently dropping
its traffic — is the worst of the three.

## Related

* #148 — the boxes this was found with.
* #150 — the Arch cloud-init failure found in the same run.

### LINKED ISSUE #155: route networks: isolation rules are applied only at define time and are lost on reboot
## Summary

The "complete isolation" of a `mode: route` network is enforced by in-memory iptables rules that are
applied **only when the network is defined**. They do not survive a host reboot or an iptables flush,
and nothing re-applies them — not `boxman up`, not network start, not autostart. The network comes back
up **unprotected**, silently, and continues to be described as isolated.

Found while measuring the blast radius of #152. Anchors against `dcde8a1`.

## Observed

Two `mode: route` networks on the same host, both active and autostarting:

| network | bridge | defined | isolation rules present |
|---|---|---|---|
| `…tiny-arch…isolated1` | `virbr13` | after last boot | **2** |
| `…tiny-ubuntu-24.04…isolated1` | `virbr7` | before last boot | **0** |

```console
$ uptime -s
2026-08-13 11:20:08

$ sudo iptables -S | grep -c -- '-i virbr7 -j DROP\|-o virbr7 -j DROP'
0
```

`virbr7`'s network is active, autostart=yes, and has a VM attached — with no isolation whatsoever. The
only difference between the two is which side of the last reboot they were created on.

## Cause

`apply_route_iptables_rule` has exactly one call site:

```
src/boxman/providers/libvirt/net.py:902   →  inside define_network()
```

Nothing else invokes it. In particular:

* `LibVirtSession.start_network()` just calls `network.start()` — no rule application.
* `up` Case 3 (all VMs running) reconciles networks, bridges and ssh-config, but never touches the
  isolation rules.
* libvirt autostart brings the network up at boot without boxman being involved at all.

And the rules are never persisted: on this host `iptables.service` is `disabled` and
`netfilter-persistent` is absent, so the ruleset is purely in-memory. Even where a persistence service
exists, boxman does not write to it — so relying on that would be accidental rather than designed.

Net effect: isolation holds until the first reboot, then quietly stops holding forever.

## Why this matters

`mode: route` is documented as providing "complete isolation … block all traffic between host and
guests in both directions". That is a security property, and it is the entire reason the mode exists —
guests on the network are supposed to be unable to reach the host. Losing it silently is worse than not
offering it, because the configuration still claims it and nothing reports otherwise.

## Suggested fix

Make isolation a reconciled property rather than a one-shot side effect of `define_network`:

1. Re-assert the rules wherever the network is brought up or reconciled — at minimum `start_network()`
   and the `up` Case 3 network-reconcile path. `_ensure_rule` is already idempotent (it probes with
   `-C` first), so re-running is cheap and safe.
2. Better: give each isolated network its own chain, e.g. `BOXMAN_ISO_<bridge>`, jumped to from
   INPUT/OUTPUT. Reconciling then means "flush chain, refill it in the intended order" instead of
   inserting individual rules and hoping the ordering survives. This also fixes the ordering fragility
   noted in #152, where a re-inserted DROP can end up above the DHCP ACCEPT and silently re-break it.
3. Consider reporting drift: if `up` finds a routed network whose rules are missing, say so rather than
   fixing it mutely — the user may want to know their isolated network has been open since the last
   reboot.

## Related

* #152 — DHCP cannot work through these same rules; a per-network chain would address both at once.

### LINKED ISSUE #156: destroy: VM survives (and destroy still reports success) when a managed save image exists
## Summary

`boxman destroy` leaves the VM defined — and still reports success — whenever the domain has a managed
save image. boxman creates those itself (suspend, and snapshots taken with memory state), so this hits
any VM that was ever suspended or snapshotted.

Anchors against `dcde8a1`.

## Observed

Destroying a sample box whose VM had a snapshot with memory state:

```console
$ boxman --conf boxes/tiny-libvirt-ubuntu-24.04-cloudinit/conf.yml destroy -y
WARNING: undefine with storage removal failed for bprj__…_boxman01
         (error: Refusing to undefine while domain managed save image exists) — retrying plain undefine
ERROR: error un-defining vm bprj__…_boxman01: Encountered a bad command exit code!
$ echo $?
0                                   # <-- success

$ virsh list --all | grep boxman01
 -  bprj__…_boxman01   shut off     # <-- still there
```

The fallback `undefine --snapshots-metadata` hits the same refusal, so both attempts fail and the
domain survives.

Worse, this compounds: the *next* `boxman up` then sees an existing VM, takes the reconcile path instead
of a full provision, and fails differently (see the companion issue about the projects-cache KeyError).
The two together leave a VM that cannot start.

## Cause

`DestroyVM.force_undefine_vm` (`src/boxman/providers/libvirt/destroy_vm.py`) never passes
`--managed-save`:

```python
result = self.virsh.execute(
    "undefine", self.name,
    "--remove-all-storage", "--wipe-storage",
    "--delete-storage-volume-snapshots", "--snapshots-metadata",
    warn=True)
if not result.ok:
    ...
    self.virsh.execute("undefine", self.name, "--snapshots-metadata")
```

libvirt refuses to undefine a domain with a managed save image unless told to remove it. Same omission
in `undefine_vm`.

Also: the fallback's result is never checked, so a failing fallback is invisible until the
`is_vm_defined()` probe at the end — and the surrounding destroy flow reports success regardless.

## Fix (implemented)

* `--managed-save` on both the storage-removal form and the plain fallback, in
  `force_undefine_vm` and `undefine_vm`.
* Check the fallback's result and log an error when it fails, rather than discovering it implicitly.

Manual recovery for an already-stuck VM:

```console
virsh managedsave-remove <domain>
virsh undefine <domain> --managed-save --snapshots-metadata --remove-all-storage
```

## Adjacent noise worth cleaning up

The same path logs `ERROR` for situations it recovered from — e.g. the storage-removal form fails with
`Storage volume 'vdb'(...) is not managed by libvirt` or `unsupported flags (0x2) in function
virStorageBackendVolDeleteLocal`, the fallback runs, and the domain is gone — yet the run ends with
`ERROR: failed to get domain '<name>'` for a domain that was correctly removed. A clean destroy looks
broken.

### LINKED ISSUE #157: destroy then up: update_network_cache KeyErrors on the unregistered project, networks never created
## Summary

After `boxman destroy`, a subsequent `boxman up` that reaches the network **reconcile** path (rather
than a full provision) cannot create the networks: `update_network_cache` raises a bare `KeyError` on
the project name, which surfaces as an opaque error and leaves the VMs unable to start.

Anchors against `dcde8a1`.

## Observed

```console
$ boxman --conf boxes/tiny-libvirt-ubuntu-24.04-cloudinit/conf.yml destroy -y   # exit 0
$ boxman --conf boxes/tiny-libvirt-ubuntu-24.04-cloudinit/conf.yml up
WARNING: network bprj__…__clstr__cluster_1__clstr__nat1 does not exist
ERROR: Error defining network: 'boxman_dev_tiny-ubuntu-24.04'      # <-- a KeyError, stringified
...
ERROR: network …nat1: failed
error: Failed to start domain 'bprj__…_boxman01'
error: Network not found: no network with matching name 'bprj__…__clstr__cluster_1__clstr__nat1'
$ echo $?
0
```

`'boxman_dev_tiny-ubuntu-24.04'` in that message is the *project name* — the tell-tale of a `KeyError`
caught by a broad `except` and formatted as if it were a message.

## Cause

`destroy` unregisters the project from `~/.config/boxman/cache/projects.json`. `update_network_cache`
(`src/boxman/providers/libvirt/net.py`) then indexes it unconditionally:

```python
project_name = self.manager.config['project']
project_cache = projects_in_cache[project_name]     # KeyError once destroy removed the key
```

`define_network` catches broad exceptions and logs `Error defining network: {exc}`, so a `KeyError`
becomes an unhelpful bare project name and the network is silently never defined.

## Why it is intermittent

It depends on which `up` branch you land in:

* **VM fully removed** → `up` Case 1 → full `provision()` → `define_networks()`, which registers the
  project first. Works.
* **VM survives destroy** (very easy to hit — see the companion `--managed-save` issue) → `up` takes
  the existing-VM reconcile path → `reconcile_networks` → `_define_network` → **KeyError**.

So the two bugs compound: a destroy that leaves the VM behind guarantees the next `up` cannot recreate
its networks, and the VM then fails to start.

## Fix (implemented)

Re-register the project entry when it is missing, rather than assuming one survived — including `conf`
and `runtime`, since other code reads those off the entry:

```python
if project_name not in projects_in_cache:
    entry = {}
    if getattr(self.manager, 'config_path', None):
        entry['conf'] = os.path.abspath(os.path.expanduser(self.manager.config_path))
    if getattr(self.manager, '_runtime_name', None):
        entry['runtime'] = self.manager._runtime_name
    projects_in_cache[project_name] = entry
```

Worth considering separately: `define_network`'s broad `except` turning a programming error into a
log line is what made this take so long to diagnose. A `KeyError` is not "the network could not be
defined for an operational reason" and probably should not be caught there at all.

mherkazandjian:
Fixes four related defects found by actually provisioning, isolating and tearing down routed networks on a real host.

## What was broken

**DHCP could never work on a `mode: route` network (#152).** DHCP terminates *on the host* — libvirt runs dnsmasq bound to the bridge address — so it travels INPUT and OUTPUT, the exact chains the isolation drops. Every `dhcp:` range declared on a routed network was dead config; the guest NIC came up link-local only. Six existing boxes carry that pattern.

**Isolation silently disappeared (#155).** The rules were applied once at define time and live only in memory. A host reboot dropped them while libvirt autostarted the network anyway. Found in the wild: two routed networks on one host, one protected and one wide open, differing only in which side of the last reboot they were created on. A lock that quietly stops locking is worse than no lock.

**`destroy` left the VM behind and still reported success (#156).** libvirt refuses to undefine a domain with a managed save image — which boxman itself creates via suspend and memory snapshots.

**`up` after `destroy` could not recreate the networks (#157).** `destroy` unregisters the project; `update_network_cache` then indexed it unconditionally and raised a bare `KeyError`, swallowed as `Error defining network: '<project>'`.

The last two compound: a destroy that leaves the VM defined sends the next `up` down the reconcile path, which is precisely the path that KeyErrored — leaving a VM that cannot start.

## Approach

Isolation moves into a chain per direction (`BXM_ISO_I_<bridge>` / `BXM_ISO_O_<bridge>`) that is flushed and refilled rather than inserted into, so ordering belongs to boxman instead of insertion history. `iptables -I` pushes to the top, so a DROP re-inserted after an ACCEPT would otherwise land above it and silently re-break DHCP.

- The DHCP hole opens **only** when the network declares `dhcp:`, so this is additive — a routed network that hands out nothing keeps exactly the block it always had.
- Teardown is unhook → flush → delete, removing the risk of apply and remove drifting apart. Pre-chain rules from older versions are migrated away on apply.
- Isolation is reconciled on every `up`/`update`, and drift is **reported**, not repaired mutely.

Opening the firewall alone made things *worse*, so the DHCP offer is trimmed too. dnsmasq advertises the bridge address as both gateway and resolver, both unreachable by construction. The router option installs a default route at metric 0 that outranks the guest's real NIC:

| | firewall fix only | with offer trimmed |
|---|---|---|
| default routes | 2 — isolated NIC winning | 1, via the real NIC |
| `ping 8.8.8.8` | 100% packet loss | 0% packet loss |

## Verification

On a real host with a live guest:

- isolation holds — bridge unreachable, tcp/53 blocked — **while** DHCP succeeds
- deleting the chains to simulate a reboot, then `boxman up`, detects, warns and repairs
- destroy leaves zero VMs, networks, chains or disk files
- lease contains an address and netmask, and no `routers` / `domain-name-servers`

## Tests

2075 passing on this branch, ruff clean. ~30 new tests: chain creation/hooking, fill order, `dhcp:` gating, direction-specific ports, legacy migration, teardown, chain-name sanitisation, shell quoting, all five drift outcomes, offer trimming per mode, `--managed-save` on all undefine forms, and project re-registration.

One pre-existing failure remains, unrelated and also present on `main`: `test_stale_file_logs_warning` asserts on `caplog` for a logger with `propagate=False`, so it can never see the record.

Closes #152
Closes #155
Closes #156
Closes #157

### DIFF
--- a/src/boxman/assets/network.xml.j2
+++ b/src/boxman/assets/network.xml.j2
@@ -1,5 +1,14 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<network>
+{%- set has_dhcp = (dhcp_range_start and dhcp_range_end) or dhcp_hosts %}
+{#- A routed network is isolated from the host by firewall rules, so the
+    gateway and resolver dnsmasq would advertise (DHCP options 3 and 6, both
+    the bridge address) are unreachable by construction. Handing them out is
+    actively harmful: the router option installs a default route at metric 0,
+    which outranks the guest's real NIC and black-holes all of its traffic.
+    An empty `dhcp-option=N` tells dnsmasq to omit that option entirely, so
+    the guest gets an address and a netmask and nothing it cannot use. #}
+{%- set trim_dhcp_options = (forward_mode == 'route') and has_dhcp %}
+<network{% if trim_dhcp_options %} xmlns:dnsmasq='http://libvirt.org/schemas/network/dnsmasq/1.0'{% endif %}>
   {# the environment has no autoescape, so every config-derived value is
      escaped explicitly -- a name holding an & or a quote would otherwise
      produce XML that libvirt rejects at net-define #}
@@ -28,4 +37,10 @@
   {% endif %}
   </ip>
 {% endif %}
+{%- if trim_dhcp_options %}
+  <dnsmasq:options>
+    <dnsmasq:option value='dhcp-option=3'/>
+    <dnsmasq:option value='dhcp-option=6'/>
+  </dnsmasq:options>
+{%- endif %}
 </network>
--- a/src/boxman/manager_parts/flows.py
+++ b/src/boxman/manager_parts/flows.py
@@ -243,6 +243,7 @@ class FlowsMixin:
                 allow_recreate=getattr(cli_args, 'recreate_networks', False),
                 auto_accept=getattr(cli_args, 'yes', False))
             self.report_network_results(network_results)
+            self.raise_on_network_failures(network_results)
 
             # a recreate power-cycles the guests attached to the network, so
             # the addresses connect_info() and the ssh config are about to be
@@ -277,9 +278,11 @@ class FlowsMixin:
         # to find the network it is wired to, and any reservation added to the
         # config since the last run has to be in dnsmasq before the guest asks
         # for a lease.
-        self.report_network_results(self.reconcile_networks(
+        network_results = self.reconcile_networks(
             allow_recreate=getattr(cli_args, 'recreate_networks', False),
-            auto_accept=getattr(cli_args, 'yes', False)))
+            auto_accept=getattr(cli_args, 'yes', False))
+        self.report_network_results(network_results)
+        self.raise_on_network_failures(network_results)
 
         # Build workdir lookup for restore operations
         vm_workdir_map = dict(self._control_vm_targets(cli_args))
--- a/src/boxman/manager_parts/networks.py
+++ b/src/boxman/manager_parts/networks.py
@@ -8,7 +8,7 @@ import os
 import time
 from typing import Any
 
-from boxman.exceptions import ConfigError
+from boxman.exceptions import ConfigError, NetworkError
 from boxman.providers.libvirt import net_reconcile
 
 
@@ -238,6 +238,22 @@ class NetworksMixin:
                     f"could not be reconnected")
             else:
                 self.logger.info(f"network {full_name}: {outcome}")
+
+    def raise_on_network_failures(self, results: dict[str, str]) -> None:
+        """
+        Turn failed networks into a failed run.
+
+        Reporting alone is not enough: a routed network that could not be
+        isolated, or one that could not be defined at all, leaves guests
+        without the connectivity -- or the containment -- the configuration
+        asked for. Continuing to exit 0 makes that look like success.
+        """
+        failed = sorted(name for name, outcome in results.items()
+                        if outcome == 'failed')
+        if failed:
+            raise NetworkError(
+                f"{len(failed)} network(s) could not be brought to the "
+                f"configured state: {', '.join(failed)}")
 
     def _vms_worth_waiting_for(self) -> list[str]:
         """
--- a/src/boxman/manager_parts/networks.py
+++ b/src/boxman/manager_parts/networks.py
@@ -154,11 +154,17 @@ class NetworksMixin:
         # do not survive a reboot or a manual flush while the network itself
         # autostarts happily without them. Re-assert them on every reconcile
         # rather than only at define time.
-        self._reconcile_network_isolation(dry_run=dry_run)
+        for full_name, outcome in self._reconcile_network_isolation(
+                dry_run=dry_run).items():
+            # an unisolated routed network is a failed network: without this
+            # `up` exits 0 while the guests can reach the host
+            if outcome == 'failed':
+                results[full_name] = 'failed'
 
         return results
 
-    def _reconcile_network_isolation(self, dry_run: bool = False) -> None:
+    def _reconcile_network_isolation(
+            self, dry_run: bool = False) -> dict[str, str]:
         """
         Re-apply the host isolation of every routed network in the config.
 
@@ -167,8 +173,9 @@ class NetworksMixin:
         view unisolated, for however long it was up. That is worth a warning
         even though it is being fixed here and now.
         """
+        outcomes: dict[str, str] = {}
         if not hasattr(self.provider, 'reconcile_network_isolation'):
-            return
+            return outcomes
 
         for cluster_name, cluster in self.config['clusters'].items():
             for network_name, network_info in (cluster.get('networks') or {}).items():
@@ -186,7 +193,9 @@ class NetworksMixin:
                 except Exception as exc:
                     self.logger.error(
                         f"network {label}: could not check isolation rules: {exc}")
+                    outcomes[full_name] = 'failed'
                     continue
+                outcomes[full_name] = outcome
 
                 if outcome == 'absent':
                     # the network is not defined; its own failure was already
@@ -206,6 +215,8 @@ class NetworksMixin:
                 elif outcome == 'failed':
                     self.logger.error(
                         f"network {label}: could not apply isolation rules")
+
+        return outcomes
 
     def report_network_results(self, results: dict[str, str]) -> None:
         """
--- a/src/boxman/manager_parts/networks.py
+++ b/src/boxman/manager_parts/networks.py
@@ -150,7 +150,62 @@ class NetworksMixin:
                         allow_recreate=allow_recreate,
                         auto_accept=auto_accept)
 
+        # Isolation rules are host iptables state, not libvirt state, so they
+        # do not survive a reboot or a manual flush while the network itself
+        # autostarts happily without them. Re-assert them on every reconcile
+        # rather than only at define time.
+        self._reconcile_network_isolation(dry_run=dry_run)
+
         return results
+
+    def _reconcile_network_isolation(self, dry_run: bool = False) -> None:
+        """
+        Re-apply the host isolation of every routed network in the config.
+
+        Reported rather than done silently: a routed network that lost its
+        rules has been reachable from the host, and from the guests' point of
+        view unisolated, for however long it was up. That is worth a warning
+        even though it is being fixed here and now.
+        """
+        if not hasattr(self.provider, 'reconcile_network_isolation'):
+            return
+
+        for cluster_name, cluster in self.config['clusters'].items():
+            for network_name, network_info in (cluster.get('networks') or {}).items():
+                if (network_info or {}).get('mode') != 'route':
+                    continue
+                label = f"{cluster_name}/{network_name}"
+                full_name = self.full_network_name(
+                    project_config=self.config,
+                    cluster_name=cluster_name,
+                    network_name=network_name)
+                try:
+                    outcome = self.provider.reconcile_network_isolation(
+                        name=full_name, info=network_info,
+                        check_only=dry_run)
+                except Exception as exc:
+                    self.logger.error(
+                        f"network {label}: could not check isolation rules: {exc}")
+                    continue
+
+                if outcome == 'absent':
+                    # the network is not defined; its own failure was already
+                    # reported by the define path
+                    self.logger.debug(
+                        f"network {label}: not defined, isolation not applicable")
+                elif outcome == 'repaired':
+                    self.logger.warning(
+                        f"network {label}: isolation rules were missing and "
+                        f"have been re-applied. A routed network is left "
+                        f"unprotected between a host reboot and the next "
+                        f"boxman run.")
+                elif outcome == 'drifted':
+                    self.logger.warning(
+                        f"[dry-run] network {label}: isolation rules are "
+                        f"missing; they would be re-applied.")
+                elif outcome == 'failed':
+                    self.logger.error(
+                        f"network {label}: could not apply isolation rules")
 
     def report_network_results(self, results: dict[str, str]) -> None:
         """
--- a/src/boxman/manager_parts/vms.py
+++ b/src/boxman/manager_parts/vms.py
@@ -815,6 +815,8 @@ class VMsMixin:
 
         self.report_network_results(network_results)
 
+        self.raise_on_network_failures(network_results)
+
         # --- categorize VMs ---
         expected_vms = set(self._get_project_vm_names())
         all_existing_vms = set(self._find_all_existing_project_vms())
--- a/src/boxman/providers/libvirt/destroy_vm.py
+++ b/src/boxman/providers/libvirt/destroy_vm.py
@@ -288,7 +288,8 @@ class DestroyVM:
         try:
             self.logger.info(f"un-defining vm {self.name}")
 
-            self.virsh.execute("undefine", self.name)
+            # see force_undefine_vm: a managed save image blocks undefine
+            self.virsh.execute("undefine", self.name, "--managed-save")
 
             # verify that the vm is no longer defined
             if not self.is_vm_defined():
@@ -329,16 +330,27 @@ class DestroyVM:
             # requires a recent libvirt, so on failure fall back to a
             # plain undefine that only drops snapshot metadata — the
             # domain must always be removable.
+            # --managed-save is not optional: libvirt refuses with "Refusing to
+            # undefine while domain managed save image exists" for any domain
+            # that was suspended or snapshotted with memory state, which
+            # boxman itself creates. Without it destroy leaves the VM defined.
             result = self.virsh.execute(
                 "undefine", self.name,
                 "--remove-all-storage", "--wipe-storage",
                 "--delete-storage-volume-snapshots", "--snapshots-metadata",
+                "--managed-save",
                 warn=True)
             if not result.ok:
                 self.logger.warning(
                     f"undefine with storage removal failed for {self.name} "
                     f"({result.stderr.strip()}) — retrying plain undefine")
-                self.virsh.execute("undefine", self.name, "--snapshots-metadata")
+                fallback = self.virsh.execute(
+                    "undefine", self.name, "--snapshots-metadata",
+                    "--managed-save", warn=True)
+                if not fallback.ok:
+                    self.logger.error(
+                        f"plain undefine also failed for {self.name}: "
+                        f"{fallback.stderr.strip()}")
 
             # verify that the vm is no longer defined
             if not self.is_vm_defined():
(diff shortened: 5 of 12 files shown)


mherkazandjian (commented):
Reviewed current head `abccd03f`; both CI jobs are green. I found four issues that should be addressed before merge:

1. **[HIGH] Existing routed networks open DHCP before their dnsmasq configuration is migrated.** `network.xml.j2` adds the empty options 3/6 only when a network is defined, but `net_reconcile.parse_network_xml()` / `desired_state()` do not model those options. Therefore an existing routed network plans as `action: none`. `_reconcile_network_isolation()` nevertheless replaces the legacy unconditional INPUT/OUTPUT drops with chains that allow DHCP. The old dnsmasq definition will still advertise the bridge as router and DNS, so the next lease installs the exact unusable default route/resolver this PR is meant to prevent. This happens on the first normal `boxman up` after upgrading, without a recreate. Please make these dnsmasq options part of reconciliation (likely structural drift if libvirt cannot update them live), and do not open the DHCP hole until the live XML is known to suppress options 3/6. A regression test should start with pre-PR routed-network XML and exercise a normal reconcile.

2. **[HIGH] `_isolation_is_intact()` validates only the INPUT/OUTPUT jump, not the chain contents.** If `BXM_ISO_*` remains hooked but is flushed or loses its terminal DROP, iptables returns from the empty/user chain and host/guest traffic is no longer isolated. Dry-run reports `ok`; a normal run refills it but returns `ok` instead of `repaired`, silently hiding the exposure. Please validate the expected DHCP exception (when applicable) and terminal DROP/order, and add an empty/wrong-content chain test.

3. **[MEDIUM] Teardown can report success while leaving the custom chains hooked.** `remove_route_iptables_rule()` removes at most one matching jump and ignores the results of `iptables -F` and `-X`. With duplicate jumps, `-D` removes one and `-X` fails because the chain is still referenced, but the method returns `True`. Remove all matching hooks (or otherwise prove uniqueness) and propagate flush/delete failures when the chain exists.

4. **[MEDIUM] Isolation-apply failures are discarded by the manager.** `_reconcile_network_isolation()` logs `failed`/exceptions but returns nothing and does not add `failed` to `reconcile_networks()`' result. Consequently `up`/`update` can continue and exit successfully while a routed network remains unisolated. Please propagate this outcome through the existing network failure path.

I did not run the suite locally because this repository requires all test tiers to run in the disposable test-runner VM; the current GitHub CI results are green on Python 3.10 and 3.12.

mherkazandjian:
All four addressed in `0e8c155`. Every finding was correct; #1 and #2 in particular were things I had reasoned about and got wrong.

**1 [HIGH] — DHCP opened before the dnsmasq config is migrated.** Confirmed, and this was the serious one. I had noted the reconcile blind spot on #152 but concluded the impact was bounded because "DHCP was impossible anyway" — missing that *this PR is what makes it possible*. On the first ordinary `up` after upgrading, an existing routed network keeps advertising the bridge as router and DNS while the new chains start permitting DHCP, producing exactly the metric-0 black hole the offer trimming exists to prevent.

Fixed by gating the hole on the **live** definition rather than the template: `_dhcp_offer_is_trimmed()` reads `net-dumpxml` and looks for `dhcp-option=3`/`6`, and `_dhcp_hole_wanted()` keeps DHCP blocked until they are present, warning with the recreate command. Covered by `test_hole_stays_shut_until_the_live_offer_is_trimmed`, which drives apply against pre-PR (untrimmed) network XML and asserts no `--dport` rule is added.

I did not make the options structural drift. That would schedule a recreate for every existing routed network and bounce the attached guests, which seemed too blunt to do implicitly — the gate makes the unmigrated state safe rather than merely detected. Happy to add drift detection on top if you'd prefer the recreate to be scheduled.

**2 [HIGH] — `_isolation_is_intact()` checked only the jump.** Confirmed. A hooked-but-empty chain returns from the user chain and isolates nothing, while the probe said `ok` and a repair returned `ok` instead of `repaired` — the drift detection defeating its own purpose. It now reads `iptables -S <chain>` and requires a terminal DROP plus the DHCP exception present exactly when this run would install one. Six cases in `TestIsolationContentsAreChecked` (correct, empty, missing DROP, unexpected hole, missing expected hole, untrimmed-offer expectations).

**3 [MEDIUM] — teardown could report success with chains still hooked.** Confirmed. Removal now loops `-D` until `-C` reports the hook gone (bounded at 16), fails loudly if it cannot, and propagates `-F`/`-X` failures — while still treating a never-created chain as a no-op. Tests cover a triple-duplicated jump and a `-X` that fails with `chain is still referenced`.

**4 [MEDIUM] — isolation failures discarded by the manager.** Confirmed. `_reconcile_network_isolation()` now returns outcomes and `reconcile_networks()` merges `failed` into its results, so an unisolated routed network travels the existing network failure path instead of letting `up` exit 0. New `tests/test_network_isolation_reconcile.py` covers failure, exception, each non-failure outcome, the repaired warning, dry-run `check_only`, and that nat networks are never probed.

Also documented the `absent` outcome, which was returned but not listed (Copilot's second comment).

2093 passing locally, ruff clean. The one failure is the pre-existing `test_stale_file_logs_warning`, also red on `main`.

mherkazandjian:
All three agreed and fixed in `457d8bd`. I verified each against the code first; all three were correct, and (1) was a case of my previous fix creating a dead end.

**1 — the migration path did not exist.** Confirmed: `reconcile_networks` returns at `if plan['action'] == 'none': continue`, and `allow_recreate` is only consulted *inside* `_recreate_network`. So `--recreate-networks` cannot act on a network that plans as `none`, and my warning was advice that could not be followed. Legacy routed networks would have stayed DHCP-blocked forever.

Last round I declined to model these options because it "would schedule a recreate for every existing routed network and bounce the attached guests". That reasoning was wrong in a way worth naming: modelling the drift does **not** force a recreate — `_recreate_network` still refuses without `allow_recreate` — it makes the drift *visible* and the migration *possible*. Declining it gave up the migration path to avoid a disruption that was never automatic.

`dhcp_options_trimmed` is now parsed from the live XML, derived in `desired_state`, and listed in `STRUCTURAL_FIELDS`. A legacy network plans as `recreate`, reports as such without the flag, and migrates with it. Tests drive real pre-PR XML through `parse_network_xml` → `diff_network` and assert `action == recreate` with the field named in `structural`; also that a partially-trimmed network (only option 3) counts as untrimmed, and that a migrated network shows no drift.

**2 — extras were not rejected.** Confirmed. Exactly the chain you describe passed:

```
-A CHAIN -j ACCEPT                          <- everything already accepted
-A CHAIN -p udp --dport 67 -j ACCEPT
-A CHAIN -j DROP
```

terminal DROP ✓, hole present ✓ → reported intact, isolating nothing. It now compares the exact rule list this run would install — same length, same order, same bodies — so extras are rejected as firmly as omissions, and the terminal DROP must be unconditional since one carrying match criteria only drops some traffic. `-m <module>` pairs are normalised away because iptables echoes back matches that were never written. Five new cases: broad ACCEPT, extra rule after the DROP, conditional DROP, wrong-port hole, and that `-m udp` normalisation is not mistaken for drift.

**3 — `failed` never reached the exit code.** Confirmed: `report_network_results()` only logs, and the two callers that read the results check for `recreated`/`partial` to decide IP waits, never for `failed`. My change made the failure visible and nothing more, while the commit message implied otherwise. New `raise_on_network_failures()` raises `NetworkError` naming every failed network, called at all three sites that report results — so `up` and `update` now abort rather than exiting 0 over an unisolated network. Tests cover the guard directly, that it is quiet when nothing failed, and the end-to-end path from a failed isolation outcome to the raise.

2107 passing, ruff clean; the lone failure is the pre-existing `test_stale_file_logs_warning`, also red on `main`. Not merging — over to you.

Issue #150, opened 2026-08-17
cloud-init: DEFAULT_NETWORK_CONFIG is netplan-only, leaves systemd-networkd guests with no network


mherkazandjian:
## Summary

`DEFAULT_NETWORK_CONFIG` is written in netplan idiom. On any guest where
cloud-init falls back to its **systemd-networkd renderer** (i.e. any distro
without netplan — Arch, and by inspection Guix/NixOS/openSUSE and friends) the
rendered config matches **no interface**, the guest never gets an address, and
template creation fails after burning the full `cloudinit_done_timeout`.

Found by actually provisioning the new `boxes/tiny-libvirt-arch-cloudinit` box
(#148). Anchors against `dcde8a1`.

## The config

`src/boxman/providers/libvirt/cloudinit_presets.py:81`

```yaml
version: 2
ethernets:
  all-en:
    match:
      name: "en*"
    dhcp4: true
  all-eth:
    match:
      name: "eth*"
    dhcp4: true
```

This relies on netplan honouring `match.name` globs. cloud-init's networkd
renderer does not: it writes the **ethernet key** into `[Match] Name=` and
discards the glob.

## Observed in the guest

Arch cloud image `Arch-Linux-x86_64-cloudimg-20260815`, probed live over the
qemu guest agent while the template VM was stuck:

```
/etc/systemd/network/10-cloud-init-all-en.network
[Match]
Name=all-en

/etc/systemd/network/10-cloud-init-all-eth.network
[Match]
Name=all-eth
```

`Name=all-en` / `Name=all-eth` match nothing. The real NIC is `eth0` (the Arch
image sets `net.ifnames=0`):

```
$ ip -brief addr      → eth0   DOWN    (no address)
$ networkctl list     → eth0   ether   off   unmanaged
```

## Failure chain

1. No `.network` unit matches → `eth0` stays **unmanaged**, no DHCP.
2. `systemd-networkd-wait-online` times out
   (`Timeout occurred while waiting for network connectivity`).
3. `cloud-final.service` is `After=network-online.target time-sync.target` and
   never starts. `systemctl list-jobs` in the stuck guest:

```
269 cloud-final.service              start waiting
174 time-sync.target                 start waiting
173 systemd-time-wait-sync.service   start running
261 sshd.service                     start waiting
113 multi-user.target                start waiting
```

4. On Arch, **both** `package_update_upgrade_install` and `scripts_user` live in
[...]

Issue #149, opened 2026-08-16
provisioning: guest config is hardwired to cloud-init, blocking declarative distros (NixOS, Guix)


mherkazandjian:
## Summary

Guest configuration is hardwired to cloud-init. Any distro that does not run
cloud-init cannot be provisioned by boxman at all — it can only be booted.

This came out of #148 (Arch / NixOS / Guix example boxes). Arch landed as a
normal cloud-init box. NixOS and Guix could only land as ISO-boot boxes, and
this issue is why.

Anchors are against `dcde8a1` (current `main`).

## The constraint

`templates:` has exactly one provisioning mechanism, and it is not optional.

`ImagesMixin.create_templates` reads only `cloudinit*` keys off the template
block (`src/boxman/manager_parts/images.py:230-237`) and hands them to
`CloudInitTemplate`. There is no alternative branch.

`CloudInitTemplate` then builds a NoCloud seed unconditionally —
`prepare_nocloud_dir` falls back to a stock payload when the box declares none
(`src/boxman/providers/libvirt/cloudinit.py:317-323`):

```python
userdata = self.cloudinit_userdata
if not userdata:
    userdata = DEFAULT_USER_DATA.format(hostname=self.template_name)
```

and the verification phase waits on the qemu guest agent and a cloud-init done
marker, with a blind `cloudinit_fallback_timeout` when the guest cannot be
polled.

So for a guest with no cloud-init the sequence is: attach a seed ISO the guest
ignores → wait for a guest agent that never answers → fall through to a blind
timeout → declare the template ready → clone VMs that **never received an SSH
key**. It fails slowly and then lies about it, which is worse than refusing.

## Who this affects

Not a NixOS/Guix curiosity — it is every declaratively-configured or immutable
guest:

| distro | image situation | cloud-init |
|---|---|---|
| NixOS | no generic cloud qcow2; only a minimal installer ISO | not enabled in any official image |
| Guix System | official qcow2 exists, but it is a GNOME desktop demo image | none at all |
| Talos | ISO / API-driven | none (config pushed by Omni) |

`boxes/talos-iso-boot` already worked around this by being an ISO-boot box with
no provisioning. #148 added two more:

- `boxes/nixos-25.05-iso-boot`
- `boxes/guix-1.5.0-iso-boot`

Three boxes now share the same shape and the same README paragraph explaining
that `boxman ssh` will not reach the guest. That repetition is the signal.

## Suggested direction

[...]

PR #159, opened 2026-08-18
feat(boxes): add Arch Linux, NixOS and Guix System examples

### LINKED ISSUE #148: boxes: add Arch Linux, NixOS and Guix System examples
## Summary

Add example boxes under `boxes/` for three distros we currently have no guest
example for: **Arch Linux**, **NixOS** and **Guix System**.

Today `boxes/` covers Ubuntu 24.04, Rocky 9, CentOS 7, Talos and TrueNAS Scale.
Every Linux box in that set is a cloud-init box built from the same template
shape, so "Debian-family or RHEL-family, driven by cloud-init" is effectively
the only guest story boxman demonstrates.

Anchors are against `dcde8a1` (current `main`).

## Why

Two separate reasons, worth keeping apart:

1. **Coverage of the cloud-init path.** Arch publishes an official cloud image
   with cloud-init preinstalled, so an Arch box is a genuinely new distro on the
   *existing* code path — a rolling-release guest with a different package
   manager and a different `os_variant`.

2. **Coverage of what is *not* the cloud-init path.** NixOS and Guix System are
   declaratively configured by design. Their native provisioning unit is a
   configuration file plus a rebuild command, not a cloud-init user-data blob.
   Writing these two boxes is the cheapest way to find out whether boxman can
   express a guest whose configuration is not cloud-init — and it probably
   cannot today (see below).

Note this is about **guests**, not hosts. `scripts/installer/check_prerequisites.py`
already classifies `arch`, `nixos` and `guix` as supported *host* families
(`_CORE_STACK["arch"]`, `_NIXOS_CORE_ADVICE`, `_GUIX_CORE_ADVICE`,
`_TOOL_PKG[...]["nixos"|"guix"]`). You can already install and run boxman on
those distros. What is missing is a box that *boots* one.

## What a box looks like today

Each box is a directory holding a `conf.yml` (some also carry a committed
`conf.rendered.yml`), e.g. `boxes/tiny-libvirt-rocky-9-cloudinit/conf.yml`:

```yaml
templates:
  template1:
    name: rocky-9.7-minimal-base-template-cloudinit
    image:
      uri: https://.../Rocky-9-GenericCloud-Base-9.7-....qcow2
      checksum: sha256:...
    disk_size: 20G
    os_variant: rocky9
    memory: 2048
    vcpus: 2
    cloudinit_done_marker: /var/log/hello-cloudinit.log
    cloudinit: |
      #cloud-config
      ...
clusters:
  cluster_1:
    base_image: rocky-9.7-minimal-base-template-cloudinit
    networks: {...}
    vms: {...}
```

The template layer assumes the image boots, runs cloud-init, writes a marker
file, and exposes a qemu-guest-agent for `boxman up` to poll.

### 1. Arch Linux — expected to be a plain box

Arch publishes an official cloud image (`Arch-Linux-x86_64-cloudimg-*.qcow2`)
with `cloud-init` and `qemu-guest-agent` available. This should drop straight
into the existing template shape.

Things to settle while writing it:

- `os_variant` — confirm what `osinfo-query os` on a current libvirt reports for
  Arch (`archlinux` is the likely id; verify rather than assume).
- Image URL and checksum. Arch images are **rolling** and the "latest" path is
  republished, so a pinned dated URL + `sha256:` is the only reproducible
  option — the same problem the Ubuntu/Rocky boxes already solve by pinning.
  Worth a comment in the box explaining it will go stale faster than the others.
- Whether `qemu-guest-agent` is present in the image or has to be installed in
  `packages:`; this determines whether `cloudinit_done_marker` polling or the
  blind `cloudinit_fallback_timeout` path is used.

### 2. NixOS — probably needs boxman work, not just a box

NixOS *can* run cloud-init (`services.cloud-init.enable = true`), but that is
not how a NixOS machine is normally configured, and a stock NixOS qcow2 does not
have it enabled. The declarative unit is `/etc/nixos/configuration.nix` plus
`nixos-rebuild switch`.

So there are two candidate designs and they should be decided before anyone
writes YAML:

- **(a) Cloud-init on NixOS.** Use or build an image with cloud-init enabled and
  reuse the existing template path verbatim. Lowest effort, produces a box that
  works, but demonstrates the least NixOS-idiomatic thing possible.
- **(b) A declarative provisioning hook.** Push a `configuration.nix` into the
  guest and run `nixos-rebuild switch`. This is what a NixOS user would expect,
  and boxman has no such hook today — the template layer's only guest-config
  mechanism is the `cloudinit:` block.

(b) is the interesting one and is arguably a separate feature request that this
box would be the first consumer of. Suggest starting with (a) so there is a
working box, and opening a follow-up for (b) if the declarative path is wanted.

### 3. Guix System — same shape of problem as NixOS

Guix System images are produced by `guix system image`, and cloud-init is not
part of the standard Guix System story at all. Same fork as NixOS: either find
or build an image that boots and is reachable over SSH with an injected key, or
teach boxman to push a Guix system declaration and reconfigure.

Guix is the most likely of the three to need a prebuilt image published
somewhere, since there is no widely-distributed "Guix cloud image" equivalent to
Arch's.

## Suggested scope

Given the above, this is really "one easy box and two design questions". Suggest
splitting on execution:

- [ ] `boxes/tiny-libvirt-arch-cloudinit/` — pinned image + checksum, cloud-init
      payload matching the Rocky/Ubuntu boxes, one network, one or two VMs.
- [ ] Decide NixOS approach (a) vs (b); land `boxes/tiny-libvirt-nixos-*/` on
      whichever is chosen, and open a follow-up issue if (b).
- [ ] Same decision and box for Guix System.

## Acceptance criteria

- Each box provisions from clean with `boxman up` and tears down with
  `boxman destroy`, on a host with no prior boxman state.
- Each box's `conf.yml` renders and parses (the `validate-box` path), and
  carries the same file set as its sibling boxes in `boxes/`.
- SSH access works through the normal cluster `admin_user` / `admin_key_name`
  mechanism — not a distro-specific side channel.
- The image URI is pinned with a `checksum:`, and any distro whose "latest"
  moves (Arch especially) says so in a comment.
- If NixOS/Guix need a non-cloud-init provisioning hook, that is raised as its
  own issue rather than special-cased inside a box.

## Open questions

- Is the point of these boxes **demonstrating boxman on more guests**, or
  **exercising the non-cloud-init provisioning path**? If it is only the former,
  option (a) everywhere is fine and this is a small task. If it is the latter,
  NixOS/Guix are blocked on a boxman feature and should be sequenced after it.
- Do we want a rolling-release guest in `boxes/` at all, given the pinned-image
  churn Arch will add?

mherkazandjian:
`boxes/` covered only Ubuntu, Rocky and CentOS — all the same cloud-init template shape. These three add a rolling-release guest and two declaratively-configured distros.

## The three boxes are not the same shape, deliberately

**`tiny-libvirt-arch-cloudinit`** — a normal template box. Provisioned and SSH-verified end to end on a real host (`ID=arch`, both NICs up, extra disk attached).

Two Arch-specific details worth reviewing:
- pins the **dated** image path rather than `images/latest/`, so the checksum stays valid — with a comment that it will 404 eventually, since Arch prunes old directories
- does a full `pacman -Syu`, because Arch does not support partial upgrades and a pinned image goes stale
- overrides `cloudinit_network_config`. Arch has no netplan, so cloud-init uses its systemd-networkd renderer, which writes the ethernet **key** into `[Match] Name=` and discards `match: {name: "en*"}` globs. boxman's default preset is keyed `all-en`/`all-eth`, so on Arch it matches nothing: no address, `cloud-final` never starts, template creation fails after the full timeout. That is #150; the box works around it explicitly.

**`nixos-25.05-iso-boot`** and **`guix-1.5.0-iso-boot`** — ISO-boot boxes, following the existing `talos-iso-boot` precedent. They boot the distro; they do **not** provision it.

That is not laziness. Neither distro has a usable path through `templates:`: NixOS publishes no generic cloud image and no official image enables cloud-init, and Guix has no cloud-init at all — its only qcow2 is a GNOME desktop demo image. boxman's template flow always builds a NoCloud seed then waits on cloud-init plus the guest agent, so either would time out and never receive an SSH key, producing a box that *looks* supported and isn't. Both READMEs state this plainly, and #149 tracks the declarative-provisioning hook they would need.

Both verified on a real host: ISO downloads and checksums, VM defined with the ISO attached, and the guest boots far enough to take a DHCP lease — NixOS reaches its `[nixos@nixos:~]$` auto-login shell, Guix reaches its guided installer.

## The Guix checksum is not upstream's

GNU publishes only a detached OpenPGP signature for that ISO, no `.sha256`. The pinned hash was computed locally **after** verifying the signature (good signature, key `A28BF40C3E551372662D14F741AAE7DCCA3D8351`, Efraim Flashner, 2026-01-22), and the README documents the full reproduce-it chain. I could not independently confirm that key against Guix's published committer keyring — savannah returned a 502 rather than a not-found — so the README presents the signature as the authority and the sha256 only as a download pin. Worth a second opinion.

## Test discovery change

`test_provision_boxes` discovers every box and asserts SSH connectivity, so it now excludes `*-iso-boot` for the same reason `*-docker-runtime` is excluded — those boxes deliberately never provision a login, so every assertion would fail by design rather than find a defect. This also fixes the two **pre-existing** iso-boot boxes, which were already being discovered and would already have failed. Also maps `archlinux` → `arch`, since osinfo uses the former and `/etc/os-release` the latter.

## Notes

2040 passing, ruff clean, all three configs pass `validate_box.py`. The one failure is pre-existing and also present on `main` (`test_stale_file_logs_warning` asserts on `caplog` for a logger with `propagate=False`).

Closes #148

### DIFF
--- a/boxes/guix-1.5.0-iso-boot/README.md
+++ b/boxes/guix-1.5.0-iso-boot/README.md
@@ -35,9 +35,10 @@ oversight:
   declaration, and that declaration can include `openssh-service-type` with
   authorized keys baked in ahead of time.
 - boxman's `templates:` path **always** builds a NoCloud seed and then waits on
-  cloud-init plus the qemu guest agent
-  (`providers/libvirt/cloudinit.py`). Either Guix artifact pushed through it
-  would time out and never receive a key.
+  cloud-init plus the qemu guest agent (`providers/libvirt/cloudinit.py`).
+  Neither published Guix artifact consumes that seed, so either would time out
+  and never receive a key. A qcow2 built with keys already declared in it would
+  boot fine — it just would not be receiving them *from boxman*.
 
 So the real boxman gaps for Guix are narrower than "no cloud-init": **dynamic
 key injection** (boxman generates a keypair at provision time, which a
--- a/boxes/guix-1.5.0-iso-boot/README.md
+++ b/boxes/guix-1.5.0-iso-boot/README.md
@@ -1,0 +1,172 @@
+# guix-1.5.0-iso-boot
+
+Boots a single VM from the official **Guix System 1.5.0 installer ISO**, using
+the same ISO-boot path as
+[`ubuntu-24.04-live-iso-boot`](../ubuntu-24.04-live-iso-boot) and
+[`nixos-25.05-iso-boot`](../nixos-25.05-iso-boot) — no template, no clone, no
+cloud-init.
+
+Verified end to end on 2026-08-17: `boxman up` exits 0 in ~2 min including the
+download, the pinned checksum matches, and the guest reaches the Guix guided
+installer's locale screen and takes a DHCP lease (hostname `gnu`).
+
+## Read this first: what this box does *not* do
+
+It **boots** Guix System's installer. It does not **install or provision** it.
+There is no admin user and no injected SSH key, so `boxman ssh` will not reach
+the guest.
+
+(boxman *does* still generate an `ssh_config` and an `id_ed25519_boxman`
+keypair in the workspace, and `boxman up` prints an `ssh …` line for the VM.
+Ignore both: the `admin` user they name does not exist in the guest. `boxman
+up` also ends with `ERROR: failed to add ssh keys to some vms`, which is the
+correct outcome here, not a fault.)
+
+As with the NixOS box, that is the current limit of the tooling rather than an
+oversight:
+
+- Guix System has **no cloud image and no cloud-init**. GNU does publish a
+  bootable `guix-system-vm-image-1.5.0.x86_64-linux.qcow2`, but it is a GNOME
+  desktop *demo* image with a fixed user — there is nothing in it that consumes
+  an injected SSH key.
+- boxman's `templates:` path **always** builds a NoCloud seed and then waits on
+  cloud-init plus the qemu guest agent
+  (`providers/libvirt/cloudinit.py`). Either Guix artifact pushed through it
+  would time out and never receive a key.
+
+Getting to a usable, SSH-able Guix guest needs a declarative provisioning hook
+in boxman — push an `operating-system` declaration and run
+`guix system reconfigure` — which does not exist yet. Until then, finishing the
+install is done in the guest console.
+
+## About the checksum
+
+This is the one box here whose checksum did **not** come from upstream. GNU
+publishes only a detached OpenPGP signature for this ISO, no `.sha256`. The
+hash pinned in [`conf.yml`](conf.yml) was computed locally *after* verifying
+that signature, and you can reproduce the whole chain:
+
+```bash
+curl -O https://ftp.gnu.org/gnu/guix/guix-system-install-1.5.0.x86_64-linux.iso
+curl -O https://ftp.gnu.org/gnu/guix/guix-system-install-1.5.0.x86_64-linux.iso.sig
+gpg --recv-keys A28BF40C3E551372662D14F741AAE7DCCA3D8351
+gpg --verify guix-system-install-1.5.0.x86_64-linux.iso.sig
+sha256sum guix-system-install-1.5.0.x86_64-linux.iso
+# 107e0a8082f03a10b15c1fb9383d2d752c1cdeda41b8db575a15550e1c2d8b4a
+```
+
+The signature is from **Efraim Flashner**, a Guix maintainer, made 2026-01-22.
+The OpenPGP signature is the real authority here; the `sha256:` in `conf.yml`
+only pins what boxman downloads so a corrupted or swapped file fails loudly.
+
+## Prerequisites
+
+- libvirt/KVM working (`virsh -c qemu:///system list`), and `sudo` for
+  `qemu:///system` (the box uses `use_sudo: true`).
+- `wget` or `curl` on `PATH` (boxman uses them to fetch the ISO).
+- **~1.13 GB** of network + disk for the ISO, plus 30 GB for the VM disk.
+- The **local** runtime — ISO boot is rejected under docker-compose, because
+  the cached ISO is not visible inside the libvirt container.
+
+## Bring it up
+
+```bash
+cd boxes/guix-1.5.0-iso-boot
+boxman up
+```
+
+> The first run downloads ~1.13 GB and boots a VM. Re-runs reuse the cached,
+> checksum-verified ISO under `~/.cache/boxman/images`.
+
+## How to test / verify
+
+Let `V=bprj__boxman_dev_guix-1.5.0-iso-boot__bprj_guix_guix01` (the
+fully-qualified libvirt name; `virsh list` shows it).
+
+**1. ISO was downloaded and checksum-verified** — the file exists in your
+configured cache dir (default `~/.cache/boxman/images`):
+
+```bash
+ls -lh <cache_dir>/guix-system-install-1.5.0-*.iso
+```
+
+Note: at default verbosity boxman prints **no** `checksum ok` line — that
+message logs at `INFO` while the default level is `STATUS`. Verification does
+run (a mismatch raises `Checksum mismatch for ISO` and evicts the file). Given
+this box's hand-computed hash, checking it yourself is worthwhile:
+
+```bash
+sha256sum <the cached iso>
+# 107e0a8082f03a10b15c1fb9383d2d752c1cdeda41b8db575a15550e1c2d8b4a
+```
+
+**2. The VM is defined and running:**
+
+```bash
+virsh -c qemu:///system list --all | grep guix01
+```
+
+**3. The ISO is attached as a CDROM:**
+
+```bash
+virsh -c qemu:///system dumpxml "$V" | grep -A2 '<boot\|cdrom\|\.iso'
+# expect: <disk device='cdrom'> … <source file='…guix-system-install-….iso'/>
+#         <boot dev='cdrom'/> <boot dev='hd'/>
+```
+
+Note the boot order you actually get is **`cdrom` then `hd`**, even though
+`IsoBootVM` asks virt-install for `hd,cdrom`: passing `--cdrom=` puts
+virt-install into two-phase install mode, and boxman never reaches the
+post-install reconfigure. For this box it makes no difference — it boots the
+installer and never completes one — but do not rely on the `hd`-first ordering.
+
+**4. It actually booted the installer** — watch the console:
+
+```bash
+virt-viewer -c qemu:///system "$V"          # or: virsh console "$V"
+```
+
+You should reach the GNU Guix **guided installer** (a text-mode/ncurses
+wizard), starting at its language-selection screen. Switching to another tty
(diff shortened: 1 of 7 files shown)


mherkazandjian (commented):
Reviewed current head `5aa3cc61`; both CI jobs are green. I found the following issues:

1. **[HIGH] The NixOS explanation is factually wrong and leads to the wrong design conclusion.** NixOS 25.05 has a native [`services.cloud-init.enable`](https://github.com/NixOS/nixpkgs/blob/nixos-25.05/nixos/modules/services/system/cloud-init.nix) module. Its official [`proxmoxImage` release job](https://github.com/NixOS/nixpkgs/blob/nixos-25.05/nixos/release.nix#L298-L310) uses [`proxmox-image.nix`](https://github.com/NixOS/nixpkgs/blob/nixos-25.05/nixos/modules/virtualisation/proxmox-image.nix#L359-L374), which enables cloud-init networking, sshd, and the qemu guest agent. The selected minimal installer ISO is not SSH-able because it is a live installer and does not consume Boxman's NoCloud seed—not because NixOS lacks cloud-init, and an in-guest `nixos-rebuild` provisioning hook is not the only possible solution. Please correct the README/config rationale and scope the actual Boxman gap accurately (for example, obtaining/building/importing a suitable cloud-init-enabled disk image).

2. **[MEDIUM] The Guix conclusion is also too strong.** Guix does not provide native cloud-init, so Boxman's current dynamic NoCloud key injection will not work unchanged. However, the official [Guix reference card](https://guix.gnu.org/guix-refcard.pdf) documents `guix system image --image-type=qcow2`, and an `operating-system` declaration can include `openssh-service-type` with authorized keys. A declaratively prebuilt image can therefore be SSH-able without a post-boot `guix system reconfigure` hook. Please distinguish the real Boxman gaps—dynamic key injection, image build/import, and guest-agent assumptions—from an inability to build an SSH-ready Guix image.

3. **[HIGH] The Arch static address is inside the declared DHCP pool.** The comment says `10.0.14.101` is outside the range, but `isolated1` declares `10.0.14.2` through `10.0.14.254`. Once #158 makes routed-network DHCP reachable, dnsmasq can lease `.101` to another guest and create a duplicate-address conflict. Move the static address outside the pool, narrow the pool, or use a MAC reservation.

4. **[MEDIUM] The new examples can silently stop being tested.** Every `*-iso-boot` box is removed from the integration parametrization and replaced with manual README steps, while the Arch URL is explicitly expected to 404 eventually and the existing fixture converts `create-templates`/`provision` failures into `pytest.skip`. That combination can leave CI green after downloads, checksums, or boot behavior have broken. Please add deterministic automated coverage for ISO config/render/download metadata and CDROM/boot XML, and make setup failures fail rather than skip for boxes selected by the integration job.

Direct answer to the SSH question: these two particular boxes are not SSH-able because they boot installer ISOs with no Boxman-created user/key. NixOS itself does support cloud-init; Guix does not natively, but it can build a qcow2 with SSH and keys declared ahead of time.

I did not run the suite locally because this repository requires all test tiers to run in the disposable test-runner VM; the current GitHub CI results are green on Python 3.10 and 3.12.

mherkazandjian:
All four addressed in `2759825`. Findings 1 and 3 were factual errors on my part; I verified both before changing anything rather than taking them on trust.

**1 [HIGH] — the NixOS explanation is wrong.** Confirmed against nixos-25.05: `proxmox-image.nix` sets `services.cloud-init` and `qemuGuest.enable`, and `release.nix` has the `proxmoxImage` job. So "no official NixOS image ships with cloud-init enabled" is simply false, and I had drawn the wrong design conclusion from it.

Corrected in both the README and the config: the box is not SSH-able because it boots a **live installer ISO** that does not consume the NoCloud seed and creates no user — a property of the artifact, not the distro. The stated boxman gap is now obtaining/building/importing a cloud-init-enabled disk image, not an in-guest `nixos-rebuild` hook.

**2 [MEDIUM] — the Guix conclusion is too strong.** Agreed. The docs now separate the real gaps — dynamic key injection (boxman generates the keypair at provision time, which a prebuilt image cannot know), image build/import, and the guest-agent assumptions in template verification — from an inability to produce an SSH-ready image, and point at `guix system image --image-type=qcow2` with `openssh-service-type` as the route that exists today.

**3 [HIGH] — the Arch static address is inside its own pool.** Confirmed: `.101` against a declared `.2`–`.254`, with a comment claiming the opposite. The pool now stops at `.99`. I also rewrote the surrounding comment, which asserted DHCP on a routed network "can never complete" — true today, but #158 makes it false, so it now points at the issue instead of baking in a soon-to-be-stale claim.

**4 [MEDIUM] — the new examples can silently stop being tested.** Agreed, and this was the gap I created. Added `tests/test_iso_boot_boxes.py`: renders and parses, requires a full `sha256:` on every ISO, rejects moving `latest` aliases, checks `cdroms` reference declared ISOs and that `boot_order`/`disk_size` are present, and asserts these clusters make no SSH promises (`admin_user`/`base_image`). Plus an opt-in `integration`-marked HEAD check that the pinned URLs still resolve. It also guards against a rename silently dropping a box from discovery.

The fixture now uses `pytest.fail` rather than `pytest.skip` for `create-templates`/`provision` failures, so a broken box turns the integration job red.

One scoping note: the README-documents-the-limitation assertion is skipped for `talos-iso-boot` and `ubuntu-24.04-live-iso-boot`. The former has no README at all and the latter predates the convention; retrofitting them felt like it belonged in its own change rather than riding along here. Say the word if you'd rather it were done now.

Also reworded the discovery docstring, which described a boot order the boxes do not declare and virt-install does not honour (Copilot's first comment).

2063 passing, 2 skipped, ruff clean; all three box configs pass `validate_box.py`.

mherkazandjian:
Both agreed and fixed in `38c790c`. These were leftovers from my previous pass, and the partial state was worse than the original: I corrected the "Read this first" prose while leaving adjacent passages asserting the opposite, so each file argued both ways.

**Guix config header** still opened with "no cloud image and no cloud-init", flatly contradicting the distinction the README had just been corrected to draw. It now separates the two facts that actually matter: this box pins the *installer* ISO, which creates no user and consumes no NoCloud seed; and Guix has no native cloud-init, so boxman's **dynamic** key injection has no mechanism to work through. Neither implies an SSH-able Guix image is impossible — `guix system image --image-type=qcow2` can declare `openssh-service-type` with authorized keys ahead of time. The gaps are named as dynamic key injection, image build/import, and the guest-agent assumptions in template verification.

**NixOS README** still said a NixOS image pushed through `templates:` would time out. True of the installer ISO pinned here, false of a cloud-init-enabled disk image — precisely the distinction the rest of the file had been corrected to make. Now scoped to the artifact, with the gap stated as "this box has no such image to point at" rather than "the path is unusable for NixOS". Applied the same scoping to the equivalent Guix sentence, which had the same problem.

I swept both boxes for other stale claims; the only remaining hits are in `conf.rendered.yml`, which is gitignored build output (`.gitignore:160`) and not part of this PR.

2063 passing, 2 skipped, ruff clean; both configs pass `validate_box.py`. Not merging — over to you.

PR #160, opened 2026-08-18
test: close the coverage gaps in PR #147's new code

mherkazandjian:
Tests only — no production code changes.

PR #147 landed **309 new runnable lines** in `src/` at **92.9%** coverage, and **82** in `scripts/installer` at **79.3%**. This closes the gaps.

| | before | after |
|---|---|---|
| `src/` new lines | 92.9% | **100%** |
| `scripts/installer` new lines | 79.3% | **96.3%** |

## Why these lines mattered

None of the 22 uncovered lines were dead code — every one was an error or rollback path. That is exactly the code least likely to be exercised by hand and most likely to be wrong, so it is the coverage worth having rather than a number worth hitting.

- **`net.py`** (the largest gap, 16 lines) — bridge flags that read back unparseable or absent; and the three rollback steps of `define_network` when *they* fail, both raising and returning falsy. Those warnings are the only place an operator learns which pieces of a half-built network were left on the host, since the return value is a bare `False` either way. Plus the managed-owner diagnostic when `net-list` is unavailable or a definition is unreadable.
- **`clone_vm.py`** — `discard_unsafe_clone()` called without a sanitizer cause must still raise, rather than quietly leaving an unsafe clone behind.
- **`commands.py`** — `virt-sysprep` `execution_timeout` rejection, bools included: `True` would otherwise render as `timeout 1s`.
- **`vms.py`** — `needs_restart` VMs are called out in `update`'s closing summary and **not** folded into the plain `updated:` list, where a change still waiting for a boot would read as finished.
- **`networks.py`** — a `ConfigError` from `plan_network` fails that one network while its siblings are still planned.
- **`check_prerequisites.py`** — the YAML backslash and doubled-quote escape branches, the non-key-line and key-absent fall-throughs, and the sudo-guidance paths that vary with the configured libvirt sudo mode.

## What is deliberately left uncovered

Three installer lines in `Doctor.__init__` and `check_local_runtime`. Both do real host probing, which is outside that file's stated "no libvirt/subprocess/host calls" scope — covering them would mean weakening that boundary for very little.

## Note

2060 passing, ruff clean. One pre-existing failure remains, unrelated to this PR and also present on `main`: `test_stale_file_logs_warning` asserts on `caplog.records` for the `boxman` logger, which sets `propagate=False`, so the root-handler-based fixture can never see the record. It fails identically under pytest 8.4.2 and 9.0.2, so it is not a version artefact. Left alone here since it is out of scope, but worth a one-line fix.

### DIFF
--- a/tests/test_check_prerequisites.py
+++ b/tests/test_check_prerequisites.py
@@ -124,6 +124,38 @@ def test_strip_yaml_comment_preserves_hashes_inside_scalars():
         "key: a#value ")
 
 
+def test_strip_yaml_comment_honours_double_quoted_backslash_escapes():
+    # the escaped quote does not close the scalar, so the '#' after it is
+    # still inside the string and must survive
+    assert checker.strip_yaml_comment('key: "a \\" # b" # comment') == (
+        'key: "a \\" # b" ')
+
+
+def test_strip_yaml_comment_honours_single_quoted_doubled_quotes():
+    # YAML escapes a single quote by doubling it; '' must not read as a
+    # close followed by a re-open
+    assert checker.strip_yaml_comment("key: 'it''s # fine' # comment") == (
+        "key: 'it''s # fine' ")
+
+
+def test_configured_libvirt_use_sudo_skips_lines_that_are_not_keys():
+    # a sequence item under the block must not be mistaken for a mapping
+    # key, and must not derail the scan for the key that follows it
+    text = (
+        "providers:\n"
+        "  libvirt:\n"
+        "    - not-a-key\n"
+        "    use_sudo: true\n"
+    )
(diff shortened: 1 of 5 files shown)


mherkazandjian (commented):
Reviewed current head `77848e7b`. I found no blocking issues.

The changes are test-only and cover the intended edge branches: YAML quote/comment parsing, `use_sudo` detection and advice, invalid `virt-sysprep` timeouts, clone-cleanup failure without a sanitizer cause, per-network validation isolation, and the `needs_restart` update summary. The diff is clean, and GitHub CI is green on Python 3.10 and 3.12.

I did not rerun tests locally because repository policy requires all test tiers to run in the disposable test-runner VM.

PR #161, opened 2026-08-19
docs: add a consolidated networking reference (doc/network.md)

mherkazandjian:
## Why

Networking was the one major subsystem with no document of its own. What existed was scattered
across three files with different audiences:

| Where | What it covered |
|---|---|
| `doc/tutorial/README.md` "Libvirt network modes" | the closest thing to user docs — nat/route/bridge, `shared_networks` vs `mode: bridge` |
| `README.md` "Advanced" | only the docker-vs-libvirt firewall conflict |
| `doc/docker-compose-provider/*` | a **different** network model (macvlan, docker nets), easy to mistake for the libvirt one |

And the routed-network isolation work was documented **nowhere**: the per-bridge
`BXM_ISO_I_*`/`BXM_ISO_O_*` chains, the DHCP exception, the dnsmasq option trimming that makes that
exception safe, and the fact that the rules are re-asserted on every `up`/`update` because they are
host iptables state a reboot removes. That is user-visible behaviour — a routed network now gets
DHCP that previously could not work.

There were also no diagrams of the network modes anywhere. The four that exist are about the
docker-compose provider, the tutorial's object graph, and the firewall conflict.

## What

`doc/network.md`, with a diagram for each major concept:

- **Choosing an attachment** — decision tree plus a table across the four models (per-cluster
  network, `shared_networks`, `is_global`), distinguishing who creates it, who destroys it, whether
  it is namespaced, and whether the guest can reach the host.
- **`nat` / `route` / `bridge`** — what libvirt does versus what boxman adds on top (nothing at all
  for `nat`), and the consequences of routed mode that bite people: no DNS from the host, no
  `boxman ssh`, no default route.
- **Namespacing** — why `nat1` in two clusters are two different networks, and the adapter
  resolution order.
- **The routed-network isolation contract** — the exact rules, why they live in dedicated chains,
  why the DHCP hole exists, why the offer is trimmed, and why the two halves are interlocked.
- **Reconciliation** — live vs structural changes, and why a recreate is disruptive.
- **`shared_networks`** and the **`is_global`** escape hatch.
- **Troubleshooting** and a map of the code.

Cross-linked from the README features list, the README firewall section, and the tutorial's
network-modes section.

## Notes for review

- Content is drawn from the code, not just relocated prose. One draft claim — that a static
  reservation must sit outside the DHCP range — turned out **not** to be enforced by
  `_parse_dhcp_hosts`, so it was replaced with the checks that are actually performed.
- All six mermaid diagrams were rendered headlessly (`@mermaid-js/mermaid-cli`) to confirm they
  parse and lay out readably; two were restructured after looking at the output.
- Every internal anchor and relative file link was verified to resolve.
- Docs only — no code changes. `pytest tests/` is green (2151 passed, 2 skipped).

### DIFF
--- a/doc/network.md
+++ b/doc/network.md
@@ -18,11 +18,19 @@ configures on the host beyond what libvirt does by itself.
 
 ## Choosing an attachment
 
-There are four ways to get a NIC into a VM, and picking the wrong one is the
-single most common networking mistake with boxman. They are not
-interchangeable: two are libvirt networks that boxman creates and destroys, one
-is a host bridge boxman creates but never destroys, and one is a reference to
-something boxman does not manage at all.
+There are five ways to get a NIC into a VM, and picking the wrong one is the
+single most common networking mistake with boxman. What separates them is who
+owns what:
+
+- **`nat`, `route` and `bridge`** are all boxman-owned libvirt networks,
+  declared per cluster, defined on provision and removed on destroy. They differ
+  in what sits *underneath*: for `nat` and `route` libvirt creates the Linux
+  bridge and deletes it again, while `bridge` wraps a host bridge that boxman
+  never creates, touches or removes.
+- **`shared_networks`** is a host Linux bridge that boxman creates but never
+  destroys.
+- **`is_global`** is a bare reference to something boxman does not manage at
+  all.
 
 ```mermaid
 flowchart TD
@@ -451,10 +459,30 @@ flowchart TB
     style BR fill:#3498db,color:#fff,stroke:none
 ```
 
-`ensure()` is idempotent and safe across projects: create if missing, bring up,
+`ensure()` is idempotent for a single declaration: create if missing, bring up,
 apply MTU and STP. There is **no teardown** — `boxman destroy` leaves shared
 bridges alone, because another project may still be using one. Removing them is
 an explicit user action.
+
+> **That "boxman will not tear it down" is the only sense in which a shared
+> bridge is safe across projects.** Bridge names are global and not namespaced,
+> and every run re-applies the settings to whatever bridge the name resolves to,
+> so two projects that declare the same bridge differently get last-run-wins.
+> Every project sharing a bridge must agree on its settings.
+
+The re-application is not uniform, which makes the disagreement easy to miss:
+
+| Setting | On every `ensure()` | If a project omits it |
+|---|---|---|
+| bridge exists, link `up` | re-applied | — |
+| `mtu` | applied only when declared | the existing MTU is left alone |
+| `stp` | **always applied** | forced **off** — the key defaults to `false` |
+| `disable_netfilter` | acts only when `true` | the host-global sysctl is **not** restored to `1` |
+
+So a project that simply does not mention `stp:` silently turns STP off for
+everyone else on that bridge, and one project's `disable_netfilter: true`
+outlives the run that set it — nothing puts `bridge-nf-call-iptables` back to
+`1` except a reboot or kubelet.
 
 By default boxman also installs a **scoped** per-bridge accept rule so bridged
 lab frames survive a docker-style `FORWARD` DROP policy:
--- a/doc/network.md
+++ b/doc/network.md
@@ -1,0 +1,542 @@
+# Networking
+
+How boxman attaches VMs to networks, what each mode does, and what boxman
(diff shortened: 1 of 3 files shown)


doc/network.md
```diff
@@ -0,0 +1,542 @@
...
+- [`shared_networks`: host bridges for hybrid labs](#shared_networks-host-bridges-for-hybrid-labs)
+- [`is_global`: attaching to a network boxman does not own](#is_global-attaching-to-a-network-boxman-does-not-own)
+- [Troubleshooting](#troubleshooting)
+- [Where the code lives](#where-the-code-lives)
+
+## Choosing an attachment
+
+There are four ways to get a NIC into a VM, and picking the wrong one is the
+single most common networking mistake with boxman. They are not
+interchangeable: two are libvirt networks that boxman creates and destroys, one
```
mherkazandjian: **Documentation accuracy:** This count/ownership description does not match the table below. There are five attachment forms shown (`nat`, `route`, `bridge`, `shared_networks`, and `is_global`), and all three per-cluster modes create a Boxman-owned libvirt network. `mode: bridge` creates/removes the libvirt network wrapper while preserving only the underlying host bridge. Please update the count and the "two are libvirt networks" wording so users do not mistake bridge mode for an unmanaged attachment.

doc/network.md
```diff
@@ -0,0 +1,542 @@
...
+    end
+    VM1["boxman VM<br/>interface type='bridge'"] --- BR
+    VM2["boxman VM<br/>(another cluster)"] --- BR
+    CL["containerlab node<br/>host:br-app endpoint"] --- BR
+    DK["docker-compose box<br/>macvlan"] --- BR
+
+    style BR fill:#3498db,color:#fff,stroke:none
+```
+
+`ensure()` is idempotent and safe across projects: create if missing, bring up,
```
mherkazandjian: **Cross-project safety:** `ensure()` is idempotent for one declaration, but this is not generally configuration-safe across projects. Since shared bridge names are global and not namespaced, every run reapplies `up`, MTU, and STP; `disable_netfilter: true` additionally changes a host-global sysctl. Two projects that declare the same bridge differently therefore get last-run-wins behavior. Please qualify "safe across projects" to mean Boxman will not tear the bridge down, and document that all projects sharing it must agree on settings.

mherkazandjian (commented):
Independent review completed against the networking implementation. I found two documentation accuracy issues to address before merge; the code paths otherwise support the guide's operational claims, and CI is green.

doc/network.md
mherkazandjian: Agreed, and fixed in 6a3b2e4. The paragraph contradicted the table directly underneath it.

Rewritten around ownership rather than a count. `nat`, `route` and `bridge` are now stated together as boxman-owned libvirt networks, with the real distinction being the layer below: libvirt creates and deletes the Linux bridge for `nat`/`route`, while `bridge` wraps a host bridge boxman never creates, touches or removes. `shared_networks` (created, never destroyed) and `is_global` (unmanaged) are listed separately, so the text now lines up with the five rows in the table.

doc/network.md
mherkazandjian: Agreed, and fixed in 6a3b2e4. "Safe across projects" was doing far too much work.

The claim is now scoped to what is actually true — boxman never tears the bridge down — with the cross-project hazard called out: names are global and not namespaced, every run re-applies settings to whatever the name resolves to, so two projects declaring the same bridge differently get last-run-wins, and all of them must agree on its settings.

Checking `ensure()` while writing that turned up something worth documenting on its own: the re-application is **not uniform**, which is exactly what makes a disagreement easy to miss.

| Setting | On every `ensure()` | If a project omits it |
|---|---|---|
| `mtu` | applied only when declared | the existing MTU is left alone |
| `stp` | always applied | forced **off** — `entry.get("stp", False)` |
| `disable_netfilter` | acts only when `true` | the host-global sysctl is **not** restored to `1` |

So the two settings fail in opposite directions: omitting `mtu` is inert, but omitting `stp` actively turns STP off for every other project on that bridge. And `disable_netfilter: true` outlives the run that set it, since nothing puts `bridge-nf-call-iptables` back to `1` except a reboot or kubelet. That is in the doc as a table now.

PR #163, opened 2026-08-20
fix(netlab): write shared-bridge stp only when a project declares it

### LINKED ISSUE #162: shared_networks: stp is written unconditionally, so a project that omits it turns STP off for every other project on the same bridge
## Summary

`shared_bridges.ensure()` applies `stp` to the host bridge on **every** run, using
`entry.get("stp", False)` as the value. Shared bridge names are global and not namespaced, so a
project that never mentions `stp:` does not leave the setting alone — it actively writes
`stp_state 0` onto a bridge another project may have deliberately enabled STP on.

`mtu`, sitting two lines above, is guarded and does the opposite: omitting it leaves the existing
value untouched. The asymmetry is what makes this easy to miss.

Found while documenting `shared_networks` for #161. Anchors against `6a3b2e4`.

## Mechanism

`src/boxman/netlab/shared_bridges.py:175-181`:

```python
if mtu is not None:                                    # <- guarded
    _run_sudo(f"ip link set dev {qbridge} mtu {mtu}")

stp = "on" if entry.get("stp", False) else "off"       # <- unguarded, defaults off
_run_sudo(f"ip link set dev {qbridge} type bridge stp_state "
          f"{1 if stp == 'on' else 0}")
```

| Setting | On every `ensure()` | If a project omits it |
|---|---|---|
| `mtu` | applied only when declared | existing MTU left alone — inert |
| `stp` | **always applied** | forced **off** |
| `disable_netfilter` | acts only when `true` | host-global sysctl **not** restored to `1` |

The history explains it rather than justifying it: the `stp` line is unchanged from the original
netlab commit `b7c6dfa`, while `mtu` was added later in `60f7fef` (#85 item 22) with an
`is not None` guard. Nothing decided that `stp` should behave differently.

## Concrete collision

Shared bridge names are reused across the shipped examples, so two projects declaring the same
bridge is the expected pattern, not an edge case:

| bridge | declared by |
|---|---|
| `clab_mgmt` | `hybrid-libvirt-containerlab-arista-ceos`, `-cisco-iosvl2`, `-nokia-srlinux` |
| `bx_inter_a` / `bx_inter_b` | `multi-cluster-routed-via-ceos`, `-frr`, `-srlinux` |

Every one of those sets `stp: false` explicitly, so nothing in-tree trips it today. It trips as soon
as one project wants STP on:

```yaml
# project A — a lab that actually exercises spanning tree
shared_networks:
  lab:
    bridge: bx_lab
    stp: true

# project B — same L2 domain, does not care about STP and so never mentions it
shared_networks:
  lab:
    bridge: bx_lab
```

`boxman up` in A enables STP. Any later `boxman up` in B disables it. Nothing warns, and A's next
run silently re-enables it — the bridge flips with whichever project ran last.

## Why this matters

STP on a shared bridge is a property of the **L2 domain**, not of the project that happens to be
running. Turning it off can create a forwarding loop across the very topology the netlab subsystem
exists to build, and it is invisible: no log line, no drift report, and `boxman up` exits 0.

It is also the one setting here that fails *closed to a value* rather than to a no-op. A user
reading the config sees no `stp:` key and reasonably concludes the project has no opinion about STP.
It has one, and it wins.

## Suggested fix

Guard `stp` the way `mtu` is guarded — apply it only when the key is present:

```python
stp = entry.get("stp")
if stp is not None:
    _run_sudo(f"ip link set dev {qbridge} type bridge stp_state {1 if stp else 0}")
```

A bridge boxman creates still needs a defined initial state, so if "absent means don't touch" is
adopted, set the default only on the create path, not on every ensure.

Worth considering alongside it: **detect the disagreement instead of silently resolving it.** Since
the hazard is two projects declaring the same bridge with different settings, and the current
behaviour is last-run-wins across all three keys, a warning when the live bridge state does not match
what this project declared would surface it without changing anyone's topology.

## Note for whoever picks this up

The current behaviour is **asserted by an existing test** —
`tests/test_netlab_shared_bridges.py:55` requires `stp_state 0` in the command list for a config that
does not declare `stp`. That assertion has to change with the fix; it is not incidental coverage that
happens to pass.

## Related

- `disable_netfilter` has the same shape and is deliberately left out of scope here: once one project
  sets `bridge-nf-call-iptables=0`, no later run restores it, and "restore it" is not obviously
  correct while another project may still need the disable. Documented in `doc/network.md` as part of
  #161; worth its own issue if it should behave differently.
- #161 — documents the current last-run-wins behaviour so it is at least not a surprise.

mherkazandjian:
Closes #162. **Stacked on #161** (base is `doc/network-guide`, not `main`) because it also corrects the
cross-project table in `doc/network.md`, which only exists on that branch. Merge #161 first and GitHub
will retarget this to `main` automatically.

## The bug

Shared bridge names are global and not namespaced, and `ensure()` wrote `stp` on **every** run using
`entry.get("stp", False)`. So a project that never mentioned `stp:` did not leave the setting alone —
it actively wrote `stp_state 0` onto a bridge a co-tenant project had deliberately enabled STP on.
Last run wins, no log line, `boxman up` exits 0.

`mtu`, two lines above, was already guarded and does the opposite: omitting it leaves the existing
value alone. The history explains the split rather than justifying it — the `stp` line is unchanged
from `b7c6dfa`, while `mtu` was added later in `60f7fef` (#85 item 22) *with* an `is not None` guard.

This is not a hypothetical collision. Bridge names are already reused across the shipped examples
(`clab_mgmt` in three containerlab boxes, `bx_inter_a`/`bx_inter_b` in three routed multi-cluster
boxes), so two projects declaring the same bridge is the expected pattern.

## The fix

`stp` is written only when the entry declares it:

```python
stp = entry.get("stp")
stp_enabled = None if stp is None else _normalise_stp(stp, entry_name)
...
if stp_enabled is not None:
    _run_sudo(f"ip link set dev {qbridge} type bridge stp_state {1 if stp_enabled else 0}")
elif created:
    _run_sudo(f"ip link set dev {qbridge} type bridge stp_state 0")
```

A bridge boxman creates in this run still gets a defined initial state (off, as before), so nothing
changes for a fresh bridge — or for any shipped example box, since they all declare `stp:` explicitly.

## Second trap on the same line

`"on" if entry.get("stp", False)` is plain truthiness, so a quoted `stp: "off"` — a non-empty and
therefore truthy string — switched STP **on**. Values now go through a normaliser accepting the same
spellings as a libvirt network's `bridge.stp` (bools, `on`/`true`/`yes`/`1`, `off`/`false`/`no`/`0`)
and rejecting anything else rather than silently reading it as off, which is what
`Network._normalise_stp` already does for the libvirt path.

Normalisation runs **before** any `ip link` call, matching `mtu`'s fail-early validation, so a typo
cannot leave a freshly created half-configured bridge behind.

## Correction to the issue

#162 claimed the old behaviour was pinned by an existing test that would have to change. That was
wrong — `tests/test_netlab_shared_bridges.py:55` asserts `stp_state 0` in the bridge-**absent** case,
which is exactly the path this fix preserves. No existing assertion needed changing; all 30 prior
tests pass untouched.

## Tests

Five added (48 total in the file, all green):

| Test | Pins |
|---|---|
| `test_stp_not_touched_when_absent_on_an_existing_bridge` | the actual fix — no `stp_state` written at all |
| `test_stp_initialised_off_on_a_bridge_boxman_creates` | the create path still gets a defined state |
| `test_stp_false_declared_is_still_applied` | an explicit `false` is an opinion and must be written |
| `test_stp_spellings_normalised` (10 cases) | quoted `"off"` does not enable STP |
| `test_invalid_stp_rejected_before_touching_the_host` | typo rejected, and rejected before any mutation |

Full suite: 2169 passed, 2 skipped. `ruff check src tests` clean.

## Out of scope

`disable_netfilter` has the same cross-project shape — once any project sets it,
`bridge-nf-call-iptables` stays at `0` until a reboot or kubelet — but "restore it" is not obviously
correct while another project may still need the disable. Left as-is and documented; worth its own
issue if it should behave differently.

### DIFF
--- a/doc/network.md
+++ b/doc/network.md
@@ -460,29 +460,35 @@ flowchart TB
 ```
 
 `ensure()` is idempotent for a single declaration: create if missing, bring up,
-apply MTU and STP. There is **no teardown** — `boxman destroy` leaves shared
+and apply MTU and STP when they are declared. There is **no teardown** — `boxman destroy` leaves shared
 bridges alone, because another project may still be using one. Removing them is
 an explicit user action.
 
 > **That "boxman will not tear it down" is the only sense in which a shared
 > bridge is safe across projects.** Bridge names are global and not namespaced,
-> and every run re-applies the settings to whatever bridge the name resolves to,
-> so two projects that declare the same bridge differently get last-run-wins.
+> and every run re-writes the settings it declares onto whatever bridge the name
+> resolves to, so two projects that declare the same bridge differently get
+> last-run-wins.
 > Every project sharing a bridge must agree on its settings.
 
-The re-application is not uniform, which makes the disagreement easy to miss:
+A setting you *do* declare is written on every run, so it is the declarations
+that collide — not the omissions:
 
 | Setting | On every `ensure()` | If a project omits it |
 |---|---|---|
 | bridge exists, link `up` | re-applied | — |
 | `mtu` | applied only when declared | the existing MTU is left alone |
-| `stp` | **always applied** | forced **off** — the key defaults to `false` |
+| `stp` | applied only when declared | left alone; a bridge boxman *creates* starts with STP off |
 | `disable_netfilter` | acts only when `true` | the host-global sysctl is **not** restored to `1` |
 
-So a project that simply does not mention `stp:` silently turns STP off for
-everyone else on that bridge, and one project's `disable_netfilter: true`
-outlives the run that set it — nothing puts `bridge-nf-call-iptables` back to
-`1` except a reboot or kubelet.
+So the way to stay out of a co-tenant's way is to leave a key out entirely.
+Declaring `stp: false` is not the same as omitting it — the first is an opinion
+that gets written over anyone who asked for `true`, the second leaves the bridge
+as it is.
+
+`disable_netfilter` is the one that cannot be undone this way: once any project
+has set it, `bridge-nf-call-iptables` stays at `0` until a reboot or kubelet puts
+it back. No later boxman run restores it.
 
 By default boxman also installs a **scoped** per-bridge accept rule so bridged
 lab frames survive a docker-style `FORWARD` DROP policy:
--- a/doc/network.md
+++ b/doc/network.md
@@ -460,7 +460,8 @@ flowchart TB
 ```
 
 `ensure()` is idempotent for a single declaration: create if missing, bring up,
-and apply MTU and STP when they are declared. There is **no teardown** — `boxman destroy` leaves shared
+apply MTU and STP when declared — plus, on a bridge it had to create, STP off as
+a defined starting point. There is **no teardown** — `boxman destroy` leaves shared
 bridges alone, because another project may still be using one. Removing them is
 an explicit user action.
 
--- a/doc/network.md
+++ b/doc/network.md
@@ -490,6 +490,12 @@ as it is.
 has set it, `bridge-nf-call-iptables` stays at `0` until a reboot or kubelet puts
 it back. No later boxman run restores it.
 
+`stp` and `disable_netfilter` accept the same spellings as a libvirt network's
+`bridge.stp` — booleans, or `on`/`true`/`yes`/`1` and `off`/`false`/`no`/`0`.
+Anything else is rejected, and a key written with no value at all is an error
+rather than a silent "leave it alone": quoting matters here, because a bare
+truthiness test would read `"off"` as on.
+
 By default boxman also installs a **scoped** per-bridge accept rule so bridged
 lab frames survive a docker-style `FORWARD` DROP policy:
 
--- a/src/boxman/netlab/shared_bridges.py
+++ b/src/boxman/netlab/shared_bridges.py
@@ -167,13 +167,14 @@ def ensure(shared_networks: dict[str, dict[str, Any]] | None) -> None:
       (``ip link set dev <br> mtu <n>``). Bridges default to 1500 while
       containerlab veth links default to 9500, so set this (e.g. 9500) on
       bridges that carry jumbo lab traffic to avoid a silent blackhole.
-    - ``stp`` (bool, optional): enable STP on the bridge. Applied only when
+    - ``stp`` (bool, or an ``on``/``off`` spelling, optional): enable STP on
+      the bridge. Applied only when
       declared. An entry that omits it leaves the current setting alone on a
       bridge that already exists, and gets STP off on one boxman creates --
       because the name is shared, so writing a default on every run would
       clobber a co-tenant project's explicit choice.
-    - ``disable_netfilter`` (bool, default **False** when the key is absent;
-      a declared value is validated like ``stp``): when False (the
+    - ``disable_netfilter`` (bool, or an ``on``/``off`` spelling; default
+      **False** when the key is absent): when False (the
       default, decision D8), lab frames are allowed by an idempotent
       per-bridge scoped ``iptables`` accept rule and the host-global
       ``bridge-nf-call-iptables`` is left untouched. When True (an explicit,
--- a/src/boxman/netlab/shared_bridges.py
+++ b/src/boxman/netlab/shared_bridges.py
@@ -49,15 +49,16 @@ def _validate_bridge_name(name: str, entry_name: str) -> None:
         )
 
 
-def _normalise_stp(value: Any, entry_name: str) -> bool:
-    """Coerce a configured ``stp`` value to a bool.
+def _normalise_bool(value: Any, entry_name: str, key: str) -> bool:
+    """Coerce a configured on/off value to a bool.
 
     Same accepted spellings as a libvirt network's ``bridge.stp`` (see
-    ``Network._normalise_stp``): yaml reads an unquoted ``stp: on`` as the
-    boolean True, while a quoted ``stp: "off"`` arrives as a string. Plain
-    truthiness is wrong for the latter -- a non-empty ``"off"`` is truthy and
-    would switch STP *on*. Anything that is neither on nor off is rejected
-    rather than quietly becoming ``off``.
+    ``Network._normalise_stp``): yaml reads an unquoted ``on`` as the boolean
+    True, while a quoted ``"off"`` arrives as a string. Plain truthiness is
+    wrong for the latter -- a non-empty ``"off"`` is truthy, so it would turn
+    the setting *on*, which for ``disable_netfilter`` means silently weakening
+    bridge filtering host-wide. Anything that is neither on nor off is
+    rejected rather than quietly becoming ``off``.
     """
     if isinstance(value, bool):
         return value
@@ -68,7 +69,7 @@ def _normalise_stp(value: Any, entry_name: str) -> bool:
     if text in ('off', 'false', 'no', '0'):
         return False
     raise ConfigError(
-        f"shared_networks[{entry_name!r}]: 'stp' must be on or off, "
+        f"shared_networks[{entry_name!r}]: {key!r} must be on or off, "
         f"got {value!r}"
     )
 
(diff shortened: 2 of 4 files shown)


src/boxman/netlab/shared_bridges.py
```diff
@@ -164,22 +192,36 @@ def ensure(shared_networks: dict[str, dict[str, Any]] | None) -> None:
                 f"integer, got {mtu!r}"
             )
 
+        # normalised before anything is touched, so a typo does not leave a
+        # freshly created, half-configured bridge behind
+        stp = entry.get("stp")
```
mherkazandjian: **Validation gap:** `entry.get("stp")` collapses an omitted key and an explicitly declared YAML `stp:` / `stp: null` to the same `None`. The explicit-null case therefore bypasses `_normalise_stp`, even though its contract says anything neither on nor off is rejected and `Network._normalise_stp(None)` does reject it. Please distinguish key presence (for example, `"stp" in entry`) from the value and add a null regression test, so only true omission means "no opinion".

src/boxman/netlab/shared_bridges.py
```diff
@@ -164,22 +192,36 @@ def ensure(shared_networks: dict[str, dict[str, Any]] | None) -> None:
...
+        # default on every run is not a no-op: a project that never mentions
+        # `stp` would switch it off for every other project sharing the
+        # bridge. Write it only when this project has an opinion -- or when
+        # boxman just created the bridge and it needs a defined initial
+        # state. `mtu` above is guarded for the same reason.
+        if stp_enabled is not None:
+            _run_sudo(f"ip link set dev {qbridge} type bridge stp_state "
+                      f"{1 if stp_enabled else 0}")
+        elif created:
+            _run_sudo(f"ip link set dev {qbridge} type bridge stp_state 0")
```
mherkazandjian, line 245: **The same truthiness hazard remains in the `disable_netfilter` branch immediately below:** `disable_netfilter: "false"` is a non-empty string, so it is treated as true and later writes `bridge-nf-call-iptables=0`, weakening filtering host-wide. That is the exact input-class bug this PR fixes for quoted `stp: "off"`, with greater impact. Please validate/normalise this value before any host mutation and cover quoted false plus invalid values; if intentionally deferred, file and link a follow-up issue rather than leaving it only as an implicit out-of-scope case.

src/boxman/netlab/shared_bridges.py
```diff
@@ -132,7 +156,11 @@ def ensure(shared_networks: dict[str, dict[str, Any]] | None) -> None:
       (``ip link set dev <br> mtu <n>``). Bridges default to 1500 while
       containerlab veth links default to 9500, so set this (e.g. 9500) on
       bridges that carry jumbo lab traffic to avoid a silent blackhole.
-    - ``stp`` (bool, default False): enable STP on the bridge.
+    - ``stp`` (bool, optional): enable STP on the bridge. Applied only when
```
mherkazandjian: **Stale cross-project contract:** the function-level docstring just above still says `Safe to call repeatedly and across projects.` This PR and `doc/network.md` establish that only non-teardown is cross-project safe: explicit MTU/STP declarations remain last-writer-wins, and `disable_netfilter` is host-global. Please qualify that sentence here as well so this API doc does not retain the claim the patch is correcting.

doc/network.md
```diff
@@ -460,29 +460,35 @@ flowchart TB
...
-The re-application is not uniform, which makes the disagreement easy to miss:
+A setting you *do* declare is written on every run, so it is the declarations
+that collide — not the omissions:
 
 | Setting | On every `ensure()` | If a project omits it |
 |---|---|---|
 | bridge exists, link `up` | re-applied | — |
 | `mtu` | applied only when declared | the existing MTU is left alone |
-| `stp` | **always applied** | forced **off** — the key defaults to `false` |
+| `stp` | applied only when declared | left alone; a bridge boxman *creates* starts with STP off |
```
mherkazandjian, line 482: **Update the existing config reference too:** `doc/docker-compose-provider/config-schema.md:143` still documents `stp: false` as `optional (default false)`. After this patch, omission preserves the live value for an existing bridge and defaults off only when Boxman creates it. Please update that user-facing schema in this PR so the two references do not prescribe different semantics.

mherkazandjian (commented):
Independent review completed against the stacked base at commit 0b70ca3. I found two input-validation hazards and two documentation contracts that need updating. These do not duplicate Copilot's two existing wording comments. `git diff --check` is clean; this PR currently has no GitHub check runs attached.

src/boxman/netlab/shared_bridges.py
mherkazandjian: Agreed, fixed in 304d901. `entry.get("stp")` returns `None` for both an absent key and an explicit `stp: null`/bare `stp:`, so a config mistake silently meant "leave the setting alone" — which after this PR is a *meaningful* behaviour, not a no-op, making the collapse worse than it was before.

All three optional keys now test membership (`"stp" in entry`), so a present-but-empty key reaches validation and raises. I extended it to `mtu` as well, which had the identical shape and which you did not flag — leaving it inconsistent immediately after fixing the other two seemed worse than the small scope increase.

Covered by `test_key_present_with_no_value_is_rejected`, parametrized over `stp`, `disable_netfilter` and `mtu`.

src/boxman/netlab/shared_bridges.py
mherkazandjian: Agreed, fixed in 304d901 — and this one is worse than the `stp` case I opened the PR for. A quoted `disable_netfilter: "false"` is a non-empty, therefore truthy, string, so a config asking for the *opposite* zeroed the host-global `bridge-nf-call-iptables`. Bridge-wide vs host-wide is the difference in blast radius.

`_normalise_stp` is now `_normalise_bool(value, entry_name, key)` and serves both keys, and both are normalised before the first `ip link` call so a typo cannot leave a freshly created half-configured bridge behind.

Tests: `test_disable_netfilter_string_false_does_not_disable` (asserts the sysctl is *not* touched and the scoped rule is installed instead), `test_disable_netfilter_truthy_spellings_disable`, and `test_invalid_disable_netfilter_rejected_before_touching_the_host`.

src/boxman/netlab/shared_bridges.py
mherkazandjian: Agreed, fixed in 304d901. I corrected the same overclaim in `doc/network.md` and left it standing in the docstring one file over.

Now narrowed to what actually holds: idempotent for a given declaration, and across projects only in the sense that boxman never tears the bridge down. It also spells out the rule this PR creates — omitting a key is how an entry stays out of a co-tenant's way, `stp: false` is an opinion and no `stp` key at all is not — and notes that `disable_netfilter` escapes even that, being a host-global sysctl nothing restores.

doc/network.md
mherkazandjian: Agreed, fixed in 304d901. `config-schema.md` said `stp: false  # optional (default false)`, which described exactly the behaviour this PR removes.

Changed to `omit to leave it as-is`, with the cross-project caveat added where the `shared_networks` block is introduced (names are global, last-run-wins on declared settings, omission is how you stay clear, `disable_netfilter` excepted). The netfilter note's `by default (\`disable_netfilter: false\`)` also now reads "absent or false", since those are no longer the same input.

`design.md:235` also has `stp: false` but makes no claim about defaults, so I left it.

src/boxman/netlab/shared_bridges.py
```diff
@@ -132,7 +156,11 @@ def ensure(shared_networks: dict[str, dict[str, Any]] | None) -> None:
...
+      declared. An entry that omits it leaves the current setting alone on a
+      bridge that already exists, and gets STP off on one boxman creates --
+      because the name is shared, so writing a default on every run would
+      clobber a co-tenant project's explicit choice.
```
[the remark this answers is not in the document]
mherkazandjian: Agreed, fixed in fb4ad95. The bullet typed it `(bool, optional)` while `_normalise_bool` accepts the same on/off spellings a libvirt network's `bridge.stp` takes, and the tests exercise strings — so the docstring alone would suggest a quoted `stp: "on"` is invalid. Both `stp` and `disable_netfilter` now say "bool, or an \`on\`/\`off\` spelling".

doc/network.md
```diff
@@ -460,29 +460,35 @@ flowchart TB
 ```
 
 `ensure()` is idempotent for a single declaration: create if missing, bring up,
-apply MTU and STP. There is **no teardown** — `boxman destroy` leaves shared
+and apply MTU and STP when they are declared. There is **no teardown** — `boxman destroy` leaves shared
 bridges alone, because another project may still be using one. Removing them is
```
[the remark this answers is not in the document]
mherkazandjian: Agreed, fixed in fb4ad95. The summary line said MTU and STP are applied "when they are declared", which drops the create path — a bridge boxman creates also gets `stp_state 0` as a defined starting point. The table below already said so, so the summary was contradicting its own section. Reworded to include it.

